PR #132, opened 2025-07-16
Automated jobs query

bcote-anl:
- Added django management command to query the qstat function and populate a database entry (used as cached)
- Added simple database entry for the qstat details
- Added django migration file
- Added feature in views.py to use cached details whenever appropriate
- Added cron job script to query the qstat function every minute


resource_server_async/management/commands/query_model_status.py
atanikan, line 1: We current use .env and the SOPHIA_QSTAT_ENDPOINT_UUID variable to query model status. If we have more than one cluster I wonder if there is a smarter way to automatically parse through them. Need not be fixed in this PR but worth thinking about.

bcote-anl:
That's a fantastic point! Amazingly, this is what we define in settings.py
ALLOWED_QSTAT_ENDPOINTS = {
    "sophia":{
        "endpoint_uuid":os.getenv("SOPHIA_QSTAT_ENDPOINT_UUID"),
        "function_uuid":os.getenv("SOPHIA_QSTAT_FUNCTION_UUID")
    }
}

This means I was able to easily loop through all clusters (here just Sophia) in the management command. I pushed the change. Please let me know what you think.

PR #145, opened 2025-09-24
added whoami route

bcote-anl:
- Route that introspects the token and return user information.
- Returns auth check error if the token is not valid.

This will create a new user in the database if not already created (in order to track access activity in the database). This URL will trigger a GET information in the database every time. The advantage is that we can now let the WebUI query this URL and let our API decide whether a user is authorized to access our service. Otherwise, we need to duplicate all of the auth checks into the webui, and we need to always keep the webui and API in sync if we change something in the API.

### DIFF
--- a/resource_server_async/views.py
+++ b/resource_server_async/views.py
@@ -8,7 +8,7 @@ import asyncio
 import time
 from django.utils import timezone
 from django.utils.text import slugify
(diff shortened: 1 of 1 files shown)


atanikan (approved):
Excellent! Love this implementation

PR #151, opened 2025-10-01
Rate limit

bcote-anl:
* Adjusting rate limit per authenticated user
* Adding request.user light object to make sure Ninja Throttles see authenticated users
* Added re-usable function to check if something is in the cache
* Bypass database operations for repeating Auth errors that are the same
NOTE: did not do it here, but we could in the future re-use the same logic for all kinds of errors.


atanikan (approved):
LGTM

PR #168, opened 2025-11-05
Auth IdP streamlined + error message edit 

bcote-anl:
- Now we do not rely on IdP UUID (which according to Globus can change over time), we only rely on the IdP domain extracted from the username field. This also helps in adding new labs in our policy, since sometime it's not obvious to find their IdP UUID, but it's very easy to just add the domain
- Modified the error message when users loose their session so that they are informed to logout from Globus and re-authenticate with "--force"


atanikan (approved):
LGTM

PR #223, opened 2026-04-08
reduced timeout and remove thread

bcote-anl:


gunicorn_asgi.config.py
```diff
@@ -31,8 +31,8 @@
 worker_class = "resource_server_async.uvicorn_workers.InferenceUvicornWorker"
 
 # Worker configuration
-workers = 5  # Reduced to 5 for our VM
-threads = 4  # Increase threads per worker for better concurrency
+workers = 5
+threads = 1
 worker_connections = 1000  # Maximum number of simultaneous clients per worker
```
masalim2: I believe this also only applies for synchronous WSGI workers: https://gunicorn.org/reference/settings/#worker_connections 

> This setting only affects the gthread, eventlet and gevent worker types.

masalim2 (approved):

PR #231, opened 2026-04-21
LCF-65: Token rate limiting

masalim2:
This PR adds token-based rate limiting to the inference service with some sane defaults that can be adjusted per-model by updating the endpoint fixtures.  

Change summary:
- Factor out the endpoint slug formatting logic
- Added `tpm_model` and `tpm_user` to the `BaseEndpoint.__init__()` contract and updated all endpoint subclasses.   This allows us to set a model-wide and per-user tokens/minute rate limit at the level of each endpoint in the system.
    - `tpm` abbreviation was chosen to align with [popular usage in LiteLLM](https://docs.litellm.ai/docs/proxy/configs#load-balancing)
    - I chose to make these non-optional in the Python layer since database defaults will flow from the Endpoint models.  I added a Django migration to set some reasonable defaults (60k per user / 100k per model) on each endpoint.  
    - This can be overriden by setting `tpm_model` and `tpm_user` in the endpoints fixture files directly.  Setting the tpm values less than or equal to 0 will completely disable the throttling.
    - The rate limiting logic is implemented on the `BaseEndpoint` class since it doesn't need to vary with the backend.  Call `endpoint.check_token_rate_limit(user)` to determine if the user has exceeded their usage of the endpoint.  Call `endpoint.record_token_usage(user, num_tokens)` to record usage every time a user consumes tokens. 
    - The views call `check_token_rate_limit` and return `HTTP 429 Too Many Requests` with context about the current usage level.
    - The Request loggers (`_upsert_request_metrics_auto` and `update_batch`) will check the `total_tokens` value and call `record_token_usage`.
    - The `TokenRateLimiter` class contains the Redis-based implementation.  TPM limiting is just disabled if the redis client is not available.  Temporary redis keys which rotate every 60 seconds are used to accumulate the total tokens at the model and user levels.  the use of redis `incrby()` and `pipeline(transaction=True)` ensures that the redis updates are atomic and safe for multiple workers.
    - Reduced `log.info` --> `log.debug` for the very high frequency cache utilities.

I think we should run the following manual tests before merging:
- [x] Regular inference OK
- [x] Streaming inference OK
- [x] Batch inference OK
- [x] SAM3 Inference OK
- [x] Regular inference request: rate limited (observed 429 response)
- [x] Batch inference request: rate limited (observed 429 response)
- [x] Rate limit clears normally after 1 minute
- [x] Rate limit can be fully disabled by setting `tpm_model` and `tpm_user` to zero in DB
- [x] Rate limit can be manually adjusted (e.g. set really low to test throttling works) in the DB


resource_server_async/rate_limiters.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def record(self, user_id: str | None, tokens: int) -> None:
+        """
+        Call after LLM response with actual token counts to record the usage of the model.
+        """
+        model_key, user_key = self._keys(user_id)
+        pipe = self.redis.pipeline(transaction=True)
+
+        pipe.incrby(model_key, tokens)
+        pipe.expire(model_key, 120)  # 2x window so stragglers don't lose the key
```
bcote-anl: I understand that the value associated with model_key is cached for 2 minutes to prevent stragglers. However, since the `check()` function above looks at `current_model < self.tpm_model)` and `current_user < self.tpm_user`, does it mean that effectively `tpm_model` and `tpm_user` are token limits per 2 minutes, not per single minute?

resource_server_async/rate_limiters.py
```diff
@@ -0,0 +1,69 @@
...
+
+        allow = True
+        if self.tpm_model > 0:
+            allow = allow and (current_model < self.tpm_model)
+        if self.tpm_user > 0:
+            allow = allow and (current_user < self.tpm_user)
+
+        return TokenLimiterCheck(allow, current_model, current_user)
+
+    def record(self, user_id: str | None, tokens: int) -> None:
```
bcote-anl, line 52: This is very cool that we can safely do that with Redis! It's simple and clean.

resource_server_async/rate_limiters.py
bcote-anl: In the `_keys()` function, I'm wondering if we should remove the `{minute}` part and simply cache the key/value for 60 seconds instead of 120 seconds? This might remove the stragglers "problem" while still preserving the nice Redis logic you have in there. Unless I'm missing something?

masalim2: - To your first point, even though the keys linger in the cache for 2 minutes, they should still "roll over" to a new key every minute as `time.monotonic() // 60` jumps up to the next value.  So the window should effectively reset every minute!  
- To your second point, yes, I like that suggestion better!  It looks like `cache.set(key, 0, nx=True, ex=60)` will do the magic trick of **only setting the TTL on the first request**.  Then I can safely call `pipe.incrby(key, tokens)` to grow the counter which will naturally expire at 60 seconds rather than trying to account for the workers clocks drifting. 

This is a cleaner approach, and I also like that the window starts at the time of the user's first request (e.g. if a user starts hitting the API at `12:23:40`, the window would extend to `12:24:40` instead of resetting at `12:24:00` and giving them a slight advantage).

bcote-anl: - To the first point, thanks for clarifying, it's a nuance I did not fully understood before your clarification, thanks!
- To the second point, I'm glad you like the idea! And it will still use the nice Redis magic you are introducing, which I like quite a lot.

resource_server_async/utils.py
```diff
@@ -621,7 +621,7 @@ def get_item_from_cache(cache_key):
     try:
         cached_item = cache.get(cache_key)
         if cached_item:
-            log.info(f"Retrieved {cache_key} from cache.")
+            log.debug(f"Retrieved {cache_key} from cache.")
```
bcote-anl, line 624: This is the best change I've seen in a while in terms of "how rewarding it is" vs "the size of the change"

resource_server_async/views.py
```diff
@@ -323,28 +321,39 @@ async def post_batch_inference(request, cluster: str, framework: str, *args, **k
...
+            request,
         )
 
     # Block access if the user is not allowed to use the endpoint
     response = endpoint.check_permission(request.auth, request.user_group_uuids)
     if (response.is_authorized == False) or response.error_message:
         return await get_response(response.error_message, response.error_code, request)
 
+    # Return 429 status if TPM limits are exceeded
+    tpm_check = endpoint.check_token_rate_limit(request.auth)
```
bcote-anl: Thanks for editing the views.py! However, here I don't think we should impose rate limit for "batch". A batch will start a specific PBS job for the user and run all of the input prompts provided by the user. It can take hours and the whole process will be triggered by one single API call. This means the user will have the PBS job (the compute node) fully dedicated to their batch request, meaning that no other user will be able to use the LLM on that node. Since the rate limit we are imposing with this PR aims to protect our resources so that our live models remain responsive, I don't think it should apply to batches.

resource_server_async/views.py
```diff
@@ -634,25 +643,33 @@ async def post_inference(
...
+    assert endpoint is not None
+
+    log.info(f"endpoint_slug: {endpoint.endpoint_slug} - user: {request.auth.username}")
 
     # Block access if the user is not allowed to use the endpoint
     response = endpoint.check_permission(request.auth, request.user_group_uuids)
     if (response.is_authorized == False) or response.error_message:
         return await get_response(response.error_message, response.error_code, request)
 
+    # Return 429 status if TPM limits are exceeded
```
bcote-anl, line 651: Perfect :)

resource_server_async/views.py
masalim2: Removing the Batch rate limiting logic!  Just echoing for my own understanding: I can see now that Batch is using a totally different function UUID that runs `python -m vllm.entrypoints.openai.run_batch` as a standalone "offline" job

bcote-anl (approved):
Looks great!

masalim2:
Closing -- this work is incorporated in #234 and will be merged there!

PR #232, opened 2026-04-21
LCF-50: Base api cluster

bcote-anl:
- Creating a new base class for clusters based on direct API access like Metis
- Updated code to use the re-usable function to make API calls through httpx
- Move Metis cluster to inherit from the direct API base cluster class

Purpose: This is to facilitate the incorporation of new upcoming clusters that are API based like Metis.


bcote-anl:
@masalim2 I ran:
make format
make lint-fix 
Was I supposed to do something else?

resource_server_async/clusters/direct_api.py
```diff
@@ -0,0 +1,119 @@
...
+        openai_endpoints: List[str],
+        allowed_globus_groups: List[str] = [],
+        allowed_domains: List[str] = [],
+        config: Dict = None,
+    ):
+        # Validate endpoint configuration
+        self.__config = ClusterConfig(**config)
+
+        # Build request headers
+        self.__headers = {
```
masalim2: Since `DirectAPIClusters` are responsible for configuring and communicating with an API, they will most likely need to store things like API keys to form the correct Authorization http headers and use a consistent `base_url` prefix. 

In these cases, I think it's a great practice to set up an [httpx Client](https://www.python-httpx.org/advanced/clients/):

```python
self.__client =  httpx.Client(base_url=base_url, headers=headers, timeout=timeout)
```

The main benefit of clients is significantly improved performance when making many frequent requests to the same server.  Under the hood, Clients create a persistent TCP connection pipe that remains open for some window of time between API requests.  This reduces OS connection waiting time and especially for HTTPS the secure handshake only has to happen once per connection.  

The secondary benefit of Clients is that that the defaults only need to be set once (like a common base URL, headers, and timeouts above) and then the subsequent requests are a bit cleaner, e.g.:

```python
response = self.__client.get("/status")
```

This could also be a great opportunity to begin introducing our own structured hierarchy of exception subclasses to streamline error handling.  We could do this by creating our own subclass of `httpx.Client` that is used throughout the application.  This can handle the various `httpx` error types and re-raise our own error classes in a way that is further handled by callers or handled at the very top by a generic [Ninja  error handler](https://django-ninja.dev/guides/errors/).  

Although error handling is really closely related topic here I don't want to derail this PR too much so perhaps we can split it to another issue as a fast followup to address the error handling!

resource_server_async/clusters/direct_api.py
```diff
@@ -0,0 +1,119 @@
...
+            allowed_globus_groups,
+            allowed_domains,
+        )
+
+    # Get formatted cluster status
+    async def get_formatted_status(self) -> SubmitHTTPXCallResponse:
+        """Fetch and return cluster status. Can be overwritten to format output."""
+
+        # Get status from cluster
+        return await submit_httpx_call(
```
masalim2: With the httpx.Client pattern, calls to the external utility function would be replaced by calls to the stateful `self.__client`

resource_server_async/clusters/direct_api.py
```diff
@@ -0,0 +1,119 @@
...
+            cluster_name,
+            cluster_adapter,
+            frameworks,
+            openai_endpoints,
+            allowed_globus_groups,
+            allowed_domains,
+        )
+
+    # Get formatted cluster status
+    async def get_formatted_status(self) -> SubmitHTTPXCallResponse:
```
masalim2: With the error handling clean up, internal method APIs like this one shouldn't have to pass around generic response objects that carry an error code, error message, and possible result.  

That makes error handling a little more verbose and plays less naturally with type checkers since `response.result` can often be `None`. 

A method like this would ideally return a specific `StatusResponse` or throw a `GetClusterStatusError` which the rest of the app can handle, or otherwise the [Ninja exception handler](https://django-ninja.dev/guides/errors/) will construct and send the right response to the user.

My mental model is this kind of pushes the HTTP Response objects to the **boundaries** of the application (where httpx interfaces with cluster APIs and where Ninja interfaces with gunicorn).  Inside those boundaries, the code simplifies through simplified data types and exceptions that can bubble up from any deeply nested function.

resource_server_async/clusters/direct_api.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+
+        # Cache the result for 60 seconds
+        try:
+            cache.set(cache_key, response, 60)
+        except Exception as e:
+            log.warning(f"Failed to cache cluster status: {e}")
+
+        # Return jobs result
+        return response
```
masalim2, line 115: Just some pseudocode to illustrate how the error refactoring might look locally.  Nothing to actually do here!

```suggestion
        # Returns the response's result or throws an error:
        jobs = await self.get_formatted_status()
       
       # Catch the specific pydantic validation error and re-raise one of our application's error types:
        try: 
           cluster_jobs = GetJobsResponse(jobs=jobs)
       except ValidationError:
           raise GetJobsError(f"Could not validate jobs from cluster status: {jobs}")
        
        # Cache the result for 60 seconds
        try:
            cache.set(cache_key, cluster_jobs, 60)
         except Exception as e:
             log.warning(f"Failed to cache cluster status: {e}")

        return cluster_jobs
```

resource_server_async/endpoints/direct_api.py
```diff
@@ -70,43 +75,25 @@ def __init__(
...
-                if response.status_code != 200:
-                    return SubmitTaskResponse(
-                        error_message=f"Error: Could not send API call to {self.config.api_url}: {response.text.strip()}",
-                        error_code=response.status_code,
-                    )
-
-                # Return result if API call worked
-                return SubmitTaskResponse(result=response.text)
+        # Submit call to API
+        httpx_response: SubmitHTTPXCallResponse = await submit_httpx_call(
```
masalim2: Direct API Endpoints could similarly use and benefit from `httpx.Clients` as the API managed clusters!

resource_server_async/utils.py
masalim2, line 35: I did notice our `utils` module is getting pretty big and might benefit from some splitting up in the future.  The issue encountering here is that many things import from `utils` while `utils` also imports from modules in `resource_server_async`, so it can lead to tricky circular import issues.  It should be possible to structure the modules in a way that the utilities are truly unaware of anything else in the package and don't have to import from it; that will help with the import DAG to avoid cycles and will probably naturally come about after we give a bit more thought to code splitting.

For example, I could imagine one module for all the AccessLog/RequestLog/BatchLog handling that depends on those models; another module might handle all of the streaming logic.

I definitely don't want to suggest all this refactoring for now, but I wanted to make a note as this module continues to grow!  Perhaps if we do adopt the `httpx.Client` pattern this would be a good opportunity to bring all of the httpx client-related utilites into a `httpx_client.py` file?

resource_server_async/utils.py
```diff
@@ -1798,3 +1800,72 @@ async def get_list_endpoints_data(request) -> GetListEndpointsDataResponse:
 
     # Return the endpoint list data
     return GetListEndpointsDataResponse(all_endpoints=all_endpoints)
+
+
+# Typing and data formatting for submit_httpx_call
+class httpx_call_methods(Enum):
```
masalim2: Let's import the Python stdlib's Enums for HTTP methods and Status codes: 

https://docs.python.org/3/library/http.html

```python
from http import HTTPMethod, HTTPStatus

assert HTTPMethod.GET == 'GET'
assert HTTPStatus.OK == 200
```

resource_server_async/utils.py
```diff
@@ -1798,3 +1800,72 @@ async def get_list_endpoints_data(request) -> GetListEndpointsDataResponse:
...
+            # Make a call and wait for the response
+            if method == httpx_call_methods.get:
+                response = await client.get(url, headers=headers)
+            elif method == httpx_call_methods.post:
+                response = await client.post(url, json=data, headers=headers)
+            else:
+                return SubmitHTTPXCallResponse(
+                    error_message=f"Error: method {method} not supported to submit with HTTPx.",
+                    error_code=500,
+                )
```
masalim2:
```suggestion
            response = await client.request(method, url, headers=headers, json=data)
```

We should be able to pass all of the `HTTPMethod` from above to client.request().  I think this is valid and should help us easily support HTTP `PUT`/`DELETE`/`PATCH` if ever needed

resource_server_async/utils.py
```diff
@@ -1798,3 +1800,72 @@ async def get_list_endpoints_data(request) -> GetListEndpointsDataResponse:
...
+                return SubmitHTTPXCallResponse(
+                    error_message=f"Error: Could not send HTTPx call to {url}: {response.text.strip()}",
+                    error_code=response.status_code,
+                )
+
+            # Return retuls if HTTPx call worked
+            return SubmitHTTPXCallResponse(result=response.json())
+
+    # Errors
+    except httpx.TimeoutException:
```
masalim2: This would be another great place for us to interpret and re-raise errors from the [httpx Error hierarchy](https://github.com/encode/httpx/blob/master/httpx/_exceptions.py#L4) into our own error hierarchy that plays nicely with the Ninja error handlers

resource_server_async/utils.py
```diff
@@ -1798,3 +1800,72 @@ async def get_list_endpoints_data(request) -> GetListEndpointsDataResponse:
...
+            elif method == httpx_call_methods.post:
+                response = await client.post(url, json=data, headers=headers)
+            else:
+                return SubmitHTTPXCallResponse(
+                    error_message=f"Error: method {method} not supported to submit with HTTPx.",
+                    error_code=500,
+                )
+
+            # Return error if something went wrong
+            if response.status_code != 200:
```
masalim2:
```suggestion
            if response.status_code >= 300:
```

We should expect some valid APIs to return `201 CREATED` on POST requests that create new resources, etc...

utils/metis_utils.py
```diff
@@ -8,14 +8,18 @@
 - Make direct API calls to Metis endpoints
```
masalim2: Do you think the content of this module could be move into methods on the `MetisCluster` and/or `MetisEndpoint` classes?  

I know that the `__init__()` methods of those are designed to be called with the configuration data from postgres, so it might be awkward to instantiate a `MetisCluster` from some place like `direct_health_monitor.py`.  However, we could still use `@staticmethod` for this purpose to group methods that make sense to live close by the Metis implementation

masalim2 (commented):
This looks great!  I'm excited to see the support extending to Minerva soon.  I put a lot of general feedback here because I'm reminded of it in reading the code.  However, this is a much broader topic on the codebase and I don't think we should do it in LCF-50.   I just wanted to illustrate my thoughts while the context is fresh!

My instinct here (and tell me if you disagree) is:

- Add a persistent httpx Client that can be used by the API Clusters/Endpoints to benefit from the connection pooling 
- Punt all the error handling to a future PR
- Do the tiny fix to accept `2xx` status codes

resource_server_async/clusters/direct_api.py
bcote-anl: I started a basic httpx_client.py file with a starting point to include HTTPx client as a self value in direct_api.py. Because Http clients are now in the class, I had to stop caching the endpoint/cluster wrapper (see edits in resource_server_async/utils.py. We are still, however, caching the database transaction to pull endpoint/cluster data from postgres. Let's figure out how to properly handle errors in your PR, I like the plan, but I did not want to over-architect it here in case the plan changes.

self.__httpx_client =  AsyncHttpClient(
            timeout=self.__config.api_request_timeout,
        )

resource_server_async/clusters/direct_api.py
bcote-anl: That's much cleaner! Now I'm doing things like:
return await self.httpx_client.get(self.config.status_url)

resource_server_async/clusters/direct_api.py
bcote-anl: I agree. Let's improve on that in another PR. For now, the get_status function now simply returns a Dict.

resource_server_async/clusters/direct_api.py
bcote-anl: That would be quite great! Let's work on that within your PR to keep the error handling work more isolated.

resource_server_async/endpoints/direct_api.py
bcote-anl: I've tried to start implementing it in endpoints/direct_api.py too :)

resource_server_async/utils.py
bcote-anl: Indeed! I've started a httpx_client.py file to get started. I removed the httpx related stuff from utils.py.

resource_server_async/utils.py
bcote-anl: Given the new httpx_client.py file, we do not need HTTPMethos anymore.

resource_server_async/utils.py
bcote-anl: Given the new httpx_client.py file, we do not need HTTPMethos anymore.

resource_server_async/utils.py
bcote-anl: Yes, let's work on that within your PR.

resource_server_async/utils.py
bcote-anl: Great catch. I implemented the change (see line 731)

utils/metis_utils.py
bcote-anl: I now migrated everything within the Metis endpoint/cluster classes. I deleted metis_utils.py. We just need to make sure we don't forget to add "status_url" in the config of the Metis cluster (it's already in the fixture repo).

resource_server_async/httpx_client.py
```diff
@@ -0,0 +1,25 @@
...
+import httpx
+
+
+class AsyncHttpClient:
+    def __init__(
+        self,
+        timeout: float = 30.0,
+        headers: dict[str, str] = {"Content-Type": "application/json"},
+    ):
+        self._client = httpx.AsyncClient(timeout=timeout, headers=headers)
```
masalim2:
```suggestion
        headers: dict[str, str] | None = None,
    ):
        if headers is None:
            headers = {"Content-Type": "application/json"}
        self._client = httpx.AsyncClient(timeout=timeout, headers=headers)
```

Small tweak to avoid all `AsyncHttpClients` sharing the same default headers dict, causing bugs if two concurrent clients are using conflicting headers.

masalim2 (approved):

PR #233, opened 2026-04-23
dashboard: separate out as its own WSGI application

lowpolyneko:
- Separates dashboard out as a separate WSGI/ASGI application.
- Temporarily disables the 10s realtime metrics polling interval.
- Adds connection pool limits for Postgres specific to the dashboard.

Fixes LCF-17.


dashboard_async/settings.py
```diff
@@ -0,0 +1,404 @@
...
+INSTALLED_APPS = [
+    "django.contrib.admin",
+    "django.contrib.auth",
+    "django.contrib.contenttypes",
+    "django.contrib.sessions",
+    "django.contrib.messages",
+    "django.contrib.staticfiles",
+    "rest_framework",
+    "resource_server",
+    "resource_server_async",
```
masalim2: Could these two apps come out?  If they're not installed, I'm wondering all the `*GLOBUS*` settings listed above could be taken out to shorten up this file!

bcote-anl: Sadly, I think the Dashboard currently use the Gateway API codes to perform some actions. For example:
        # Get the jobs response from the cluster wrapper
        wrapper_response: ClusterWrapperResponse = async_to_sync(get_cluster_wrapper)(
            cluster
        )
        if wrapper_response.cluster:
            jobs_response: GetJobsResponse = async_to_sync(
                wrapper_response.cluster.get_jobs
            )(mock_auth)
            err: str = jobs_response.error_message
            cluster_status: Jobs = jobs_response.jobs
        else:
            err: str = wrapper_response.error_message
            cluster_status = None
So, either we fully decouple the dashboard and use sqlalchemy (or something else) to query Postgres and use a service account Globus client to contact our endpoints, or we temporarily run the whole thing as a separate gunicorn application to bring back the dashboard quicker.

lowpolyneko: Correct. That being said, solong as the endpoints of the resource server themselves aren't included in the list of URLs, the dashboard should only rely on the application for the model definitions.

I think `resource_server` can be pruned, however. The `_async` version is the only one the dashboard views depend on.

bcote-anl:
Looks good to me as a starting point to bring the dashboard back online! I bet we would simply create a new systemd service for it. We can probably just reuse something similar to the gateway_async.service file. But limit to 1 gunicorn worker. @lowpolyneko Would you mind adding a dashboard.service file to this PR? Thanks!

lowpolyneko:
> Looks good to me as a starting point to bring the dashboard back online! I bet we would simply create a new systemd service for it. We can probably just reuse something similar to the gateway_async.service file. But limit to 1 gunicorn worker. @lowpolyneko Would you mind adding a dashboard.service file to this PR? Thanks!

Done. I'll attempt to deploy this feature branch on Monday and, if successful, will update the `nginx.conf` aswell and merge this in.

bcote-anl (approved):

PR #235, opened 2026-04-23
LCF 50: Dynamic globus credentials

bcote-anl:
Adding optional configuration fields in the Globus Compute endpoint adaptor:
- client_id_env_name (None by default)
- client_secret_env_name (None by default)

Those are environment variables that we can add to the `.env` file to overwrite the Globus Compute endpoint credentials used for a given endpoint/model. This will be particularly useful for load testing when we do not want to interfere with the production Globus client running the endpoint. It could also be useful if we end up having multiple clusters with Globus Compute endpoints. It might be more elegant to have separate Globus credentials for different clusters.

### DIFF
--- a/inference_gateway/settings.py
+++ b/inference_gateway/settings.py
@@ -15,9 +15,9 @@ import os
 from pathlib import Path
 
 from dotenv import load_dotenv
+from pydantic import TypeAdapter
 
 from inference_gateway.utils import (
-    BackendUtilsError,
     GlobusCredentials,
     textfield_to_strlist,
 )
@@ -48,17 +48,10 @@ GLOBUS_GROUP_MANAGER_ID = os.getenv("GLOBUS_GROUP_MANAGER_ID", "")
 GLOBUS_GROUP_MANAGER_SECRET = os.getenv("GLOBUS_GROUP_MANAGER_SECRET", "")
 
 # Overwrite Globus Compute credentials (ID/secret) for specific endpoint UUIDs
-credentials_overwrite = json.loads(
-    os.getenv("GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES", "{}")
+ta = TypeAdapter(dict[str, GlobusCredentials])
+GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES = ta.validate_json(
+    os.environ.get("GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES", "{}")
 )
-try:
-    GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES = {
-        key: GlobusCredentials(**value) for key, value in credentials_overwrite.items()
-    }
-except Exception as e:
-    raise BackendUtilsError(
-        f"Could not load GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES into a dictionary. {e}"
-    )
 
 # Globus Dashboard Application Credentials
 GLOBUS_DASHBOARD_APPLICATION_ID = os.getenv("GLOBUS_DASHBOARD_APPLICATION_ID")
--- a/inference_gateway/settings.py
+++ b/inference_gateway/settings.py
@@ -16,7 +16,11 @@ from pathlib import Path
 
 from dotenv import load_dotenv
 
-from inference_gateway.utils import textfield_to_strlist, BackendUtilsError, GlobusCredentials
+from inference_gateway.utils import (
+    BackendUtilsError,
+    GlobusCredentials,
+    textfield_to_strlist,
+)
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = Path(__file__).resolve().parent.parent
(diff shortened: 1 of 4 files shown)


masalim2:
This is a great degree of extra freedom!  Just thinking about how this will be used as an admin:

On the endpoint side:
- I create a new Globus client secret
- I store `MY_ENDPOINT_CLIENT_ID` and `MY_ENDPOINT_CLIENT_SECRET` in a new file like `my_endpoint.env`
- I run `source ~/my_endpoint.env` before starting the endpoint

On the gateway side:
1.  I add `client_id_env_name="MY_ENDPOINT_CLIENT_ID"` and `client_secret_env_name="MY_ENDPOINT_CLIENT_SECRET"` to the endpoint fixture and load it with `manage.py loaddata`
2.  I add `export MY_ENDPOINT_CLIENT_ID=...` and `export MY_ENDPOINT_CLIENT_SECRET=...` to the .env of the gateway and reload the server so it picks up the matching key.

That is pretty nice!  Do you think it would be worth it to save a step if we solely read the keys from the environment?  For example:

```bash
export ENDPOINT_CREDENTIAL_OVERRIDES='
    {
        "<ENDPOINT1_UUID>": {"client_id": "<CLIENT_ID1>", "client_secret": "<CLIENT_SECRET1>"},
        "<ENDPOINT2_UUID>": {"client_id": "<CLIENT_ID2>", "client_secret": "<CLIENT_SECRET2>"},
}'
```

The reason being that we could skip Step 1 above and just set the environment/reload, and the credential configs wouldn't need to be maintained in the fixtures repo.  Let me know what you think, I'm fine with either way!

bcote-anl:
That is a great plan Misha. Let me implement that.

bcote-anl:
@masalim2 One thing I'm starting to worry about this plan is that the get_compute_client_from_globus_app function is cached to make sure we don't create too many Globus Client all the time. The get_compute_executor function is also cached based on the input client, so I have the feeling we might start creating many Client and Executors, which has been a problem in the past with Globus. The performance was degrading when too many Executors were created under the same credentials. 

I think I have a plan.. I'll make a new commit with a proposed fix

inference_gateway/settings.py
```diff
@@ -43,6 +47,18 @@
...
+    os.getenv("GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES", "{}")
+)
+try:
+    GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES = {
+        key: GlobusCredentials(**value) for key, value in credentials_overwrite.items()
+    }
+except Exception as e:
+    raise BackendUtilsError(
+        f"Could not load GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES into a dictionary. {e}"
+    )
```
masalim2: Since we're already using a BaseModel, we might as well leverage `pydantic.TypeAdapter` to validate the full settings object in one shot while guaranteeing it carries the full, accurate typing information (`dict[str, GlobusCredentials]`).  

```suggestion
ta = TypeAdapter(dict[str, GlobusCredentials])
GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES = ta.validate_json(os.environ.get("GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES", '{}'))
```

Since the settings module is loaded on startup and any exception should immediately crash the program and tell us what's wrong, I think it's okay to let the pydantic `ValidationError` crash the program rather than catch and re-raise a custom error type, but it's up to you!

utils/globus_utils.py
```diff
@@ -22,10 +23,48 @@ class ResourceServerError(Exception):
 executor_cache = TTLCache(maxsize=1024, ttl=60 * 10)
 
 
+# Get authenticated Compute Client from endpoint ID
+def get_compute_client_from_endpoint_id(endpoint_id: str = None) -> Client:
```
masalim2:
```suggestion
def get_compute_client_from_endpoint_id(endpoint_id: str) -> Client:
```

Since we already have  `get_compute_client_from_globus_app` that will build a default client, we can avoid setting a default here, so the usage is simpler: 

- Callers of `get_compute_client_from_endpoint_id` **must** provide a specific endpoint
- Callers of `get_compute_client_from_globus_app` may provide credentials or get the default endpoint

utils/globus_utils.py
```diff
@@ -22,10 +23,48 @@ class ResourceServerError(Exception):
...
+            client_id = settings.GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES[
+                endpoint_id
+            ].client_id
+            client_secret = settings.GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES[
+                endpoint_id
+            ].client_secret
+        except Exception:
+            raise ResourceServerError(
+                f"Could not retrieve Globus credentials from environment for endpoint {endpoint_id}."
+            )
```
masalim2:
```suggestion
    if (credentials := settings.GLOBUS_ENDPOINT_CREDENTIALS_OVERRIDES.get(endpoint_id)):
        client_id = credentials.client_id
        client_secret = credentials.client_secret
```

One benefit of validating the settings and forcing a specific `endpoint_id` is that this code should be able to simplify like above.  There shouldn't be any exception that can be thrown, so we can remove the guarding block here.

masalim2 (approved):
This

masalim2:
> @masalim2 One thing I'm starting to worry about this plan is that the get_compute_client_from_globus_app function is cached to make sure we don't create too many Globus Client all the time. The get_compute_executor function is also cached based on the input client, so I have the feeling we might start creating many Client and Executors, which has been a problem in the past with Globus. The performance was degrading when too many Executors were created under the same credentials.
> 
> I think I have a plan.. I'll make a new commit with a proposed fix

@bcote-anl I think the Globus Compute Clients seem to be fairly cheap, from looking at their source it's should just be a passive http client that doesn't consume too many resources when idling.  However, I see that Executors spawn background threads to actively batch tasks and push them to the service and stuff like that.  I agree we should definitely limit the number of live executors for that reason!  

 I saw that we have an in-memory cache for executors.  Can we mitigate the problem just by shrinking the cache size to a reasonable max (e.g reduce maxsize here to ~20)?

```python
# Define separate cache object for Globus executor
executor_cache = TTLCache(maxsize=1024, ttl=60 * 10)
```

bcote-anl:
Hi @masalim2 , regarding the cache maxsize for the executor, the issue was more on the Globus service side. We ran benchmarks before, and when I had several Executors authenticated with the same credentials, the performance (rate of Globus task submission) became pretty bad.

inference_gateway/settings.py
bcote-anl: Great plan, thanks!
Done.

utils/globus_utils.py
bcote-anl: Absolutely! Nice catch.
Done.

utils/globus_utils.py
bcote-anl: That's very clean. I like it.
Done.

PR #236, opened 2026-05-01
deploy: update `.service` and `nginx.conf` examples

lowpolyneko:
Replicates a subset of the configuration on our production instance.


deploy/nginx/nginx_app.conf
bcote-anl, line 1: Thanks for decoupling the application!
How does this relate this the Nginx file we are using?
/etc/nginx/sites-available/default

Two points I would like to discuss are:

Should we remove the /dashboard URL in the default file?
We are expected to use HTTPS (port 443) everywhere possible. Would it be possible to update the "listen" part? You can reuse the same SSL certificate as in the default file.

deploy/gateway_async.service
bcote-anl, line 1: Is this to make sure our API comes back online when we reboot the VM?

deploy/nginx/nginx_app.conf
lowpolyneko: The initial version was intended to be a subset of the production nginx deployment for users to utilize as a template and tailor accordingly. That being said, there is value in placing the SSL configuration, the maintenance redirect, and the rate limit configuration within the sample to encourage good practice from the user. I've updated my PR to reflect that.

I think it's okay to leave the `/dashboard` endpoint as is; users can just comment it out if they don't intend to deploy the dashboard.

bcote-anl (approved):
Looks great! Thanks

deploy/gateway_async.service
lowpolyneko: The `Restart=` coerces systemd to attempt to restart the service if it abruptly exits (i.e. the gateway encounters an uncaught exception). It's here to ensure events like these won't cause the inference service to go down.

PR #238, opened 2026-05-06
resource_server/tests: use `unittest.mock` for mocking

lowpolyneko:
Avoids hacky monkey patching of module-level globals.


bcote-anl (approved):
Much cleaner. Thanks!

PR #239, opened 2026-05-07
added status route for publicly-available clusters

bcote-anl:
to be able to report up/down/degraded for the IRI Facility Status API

### DIFF
--- a/resource_server_async/views/core.py
+++ b/resource_server_async/views/core.py
@@ -46,10 +46,9 @@ async def status_check(request: HttpRequest) -> dict[str, bool]:
     user_group_uuids = []
 
     # Get list of all publicy-available clusters
-    db_clusters = [c async for c in Cluster.objects.all()]
     authorized_clusters = [
         c
-        for db_cluster in db_clusters
+        async for db_cluster in Cluster.objects.all()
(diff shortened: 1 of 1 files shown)


resource_server_async/views/core.py
```diff
@@ -29,6 +28,39 @@ async def health_check(request: HttpRequest) -> dict[str, str]:
...
+        "idp_name": f"{prefix}-idp-name",
+    }
+    user = User(**mock_user_data)
+    user_group_uuids = []
+
+    # Get list of all publicy-available clusters
+    db_clusters = [c async for c in Cluster.objects.all()]
+    authorized_clusters = [
+        c
+        for db_cluster in db_clusters
```
lowpolyneko: It'd probably be better to `async for` all `Cluster` objects whilst collecting the `authorized_clusters`. Other requests would be able to utilize the database while the adapter is being loaded and permissions are being checked instead of greedily collecting all the cluster objects in one-shot.

Otherwise looks good! :)

bcote-anl: Thanks for the feedback! I've updated the lines. Could you have a look?

lowpolyneko (approved):
LGTM

PR #244, opened 2026-05-13
Fix health monitor

masalim2:
image

Updated the health checker to work with the new structure.  Tested in Dev and working as expected


bcote-anl (approved):
Thanks for the fix!

PR #247, opened 2026-05-14
Fix torn writes when logging large prompts/results.

masalim2:
- Update rotation logic to daily with `out.log` symlink and timestamped file suffixes. Files will only rotate once at midnight, regardless of size.  This alone greatly reduces the probability that a very large logging event will get torn between two files on rotation.  And `tail -f out.log` will now always correctly follow the latest log file, even after a rotation at midnight.
- Truncate long prompts/results before serialization. Set ~1800 per field limit.  Logic is that 2 large fields (prompt+result) on RequestLog will still fit within the 4k PIPE_BUF size limit on linux where atomic writes are guaranteed.
- Save large prompt/result pairs to {PROMPT_STORAGE_DIR}/{request_id}.json
- Change write_logs() logic to run under sync_to_async() threadpool to reduce impact of slow logging/prompt dumping on the event loop.

### DIFF
--- a/resource_server_async/logging.py
+++ b/resource_server_async/logging.py
@@ -62,7 +62,8 @@ def initialize_access_log(request: HttpRequest) -> AccessLogPydantic:
     )
 
 
-def _write_logs_sync(
+@sync_to_async(thread_sensitive=False)
+def write_logs(
     context: RequestContext, response: HttpResponse | StreamingHttpResponse
 ) -> None:
     context.access_log.emit(context.user, response)
@@ -77,9 +78,6 @@ def _write_logs_sync(
 
         if not isinstance(response, StreamingHttpResponse):
             async_to_sync(context.request_log.emit_metrics)()
-
-
-write_logs = sync_to_async(_write_logs_sync, thread_sensitive=False)
 
 
 class AccessLogMiddleware:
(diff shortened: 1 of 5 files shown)


resource_server_async/logging.py
```diff
@@ -76,7 +76,10 @@ async def write_logs(
         context.request_log.emit(body, response.status_code)
 
         if not isinstance(response, StreamingHttpResponse):
-            await context.request_log.emit_metrics()
+            async_to_sync(context.request_log.emit_metrics)()
+
+
+write_logs = sync_to_async(_write_logs_sync, thread_sensitive=False)
```
lowpolyneko: If you don't ever intend on exposing `_write_logs_sync`, you can use `@sync_to_async(thread_sensitive=False)` as a function decorator!

lowpolyneko (commented):
LGTM otherwise though

masalim2:
I tested by sending in a very large prompt (150k character content of alice in wonderland) and observing that the the log is correctly truncated.  The log line is now only ~2000 chars for this huge payload:

image

Furthermore, the prompts are retained at the configured directory (`PROMPT_STORAGE_DIR` defaults to `./prompt-records`) 

image

and contain the full `{prompt, result}` pair for each request where a prompt OR requested was truncated.  The file name is just the `{request_log.id}.json`

resource_server_async/logging.py
masalim2: Thank you, great suggestion!

bcote-anl (approved):

PR #248, opened 2026-05-15
Slight log enhancements

masalim2:
- alcf-ai client: print chat streaming errors more clearly to user
- Drop `-l` (local) flag from rotatelogs, so that rotation happens at midnight UTC instead of midnight Chicago time.  This will keep the timestamps in the log lines aligned with the file names.
- Add `-c` (create) flag to rotatelogs, so that a new empty file is always created at midnight instead of waiting for the first log line to arrive.


lowpolyneko (approved):
LGTM

PR #249, opened 2026-05-20
LCF-13 added reusable env setup and function to simplify worker_init

bcote-anl:
### Overview

First attempt at generalizing our Globus Compute endpoint worker_init to 1) ease the maintenance, and 2) help with the migration towards using templates.

### This PR includes
- `endpoint_startup_environment.sh` which defines env variables and re-usable function to streamline worker_inits
- Example of how to launch the openai oss bundle and the gemma-4-31b-it model


masalim2 (approved):

lowpolyneko (approved):
LGTM

PR #250, opened 2026-05-27
globus_utils: shutdown `Executor` on timeout

lowpolyneko:
Leaving `Executor`s alive after a timed-out request will cause the backing `ResultWatcher` thread polling on behalf of the cancelled future to leak for the duration of the program's lifetime.

The remaining threads will be `join()`-ed prior to `exit()`, meaning requests which raise a `TimeoutError` will cause the program to hang at program exit unless `Executor.shutdown(wait=False)` is called prior.

Fixes LCF-76.

### DIFF
--- a/resource_server_async/globus_utils.py
+++ b/resource_server_async/globus_utils.py
@@ -2,7 +2,6 @@ import ast
 import asyncio
 import json
 import logging
-import time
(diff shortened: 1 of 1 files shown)


bcote-anl (approved):
Thanks!

PR #252, opened 2026-06-01
LCF-109 Adding route for model details

bcote-anl:
Adding route to be able to query model details, including token rate limits (if any) and serving parameters like
    "max_model_len"
    "max_num_seqs"
    "enable_auto_tool_choice"
    "tool_call_parser"

### DIFF
--- a/resource_server_async/services.py
+++ b/resource_server_async/services.py
@@ -67,7 +67,7 @@ async def get_all_endpoints(
     """Generate and return all endpoint adaptors for a given cluster."""
 
     # For all endpoints in the database for this cluster ...
-    endpoint_adaptors: BaseEndpoint = []
+    endpoint_adaptors: List[BaseEndpoint] = []
     async for db_endpoint in Endpoint.objects.filter(cluster=cluster.cluster_name):
         endpoint = await BaseEndpoint.load_adapter(
             db_endpoint.cluster, db_endpoint.framework, db_endpoint.model
--- a/resource_server_async/services.py
+++ b/resource_server_async/services.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import uuid
-from typing import Any
+from typing import Any, List
 
 from django.conf import settings
 from django.http import StreamingHttpResponse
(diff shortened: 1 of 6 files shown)


resource_server_async/views/core.py
```diff
@@ -23,14 +23,14 @@
 # Health Check (GET) - No authentication required
 # Lightweight endpoint for Kubernetes/load balancer health checks
 @router.get("/health", auth=None)
-async def health_check(request: HttpRequest) -> dict[str, str]:
+async def health_check(request: HttpRequest) -> Dict[str, str]:
```
lowpolyneko: I believe `dict[str, str]` is the canonical type hint; `Dict` is legacy.

lowpolyneko (approved):
Minor nit on the typing, but otherwise LGTM!

resource_server_async/services.py
```diff
@@ -61,6 +61,26 @@
 logger = logging.getLogger(__name__)
 
 
+async def get_all_endpoints(
+    user: UserPydantic, cluster: BaseCluster
+) -> List[BaseEndpoint]:
```
masalim2: Same here as for the dict type annotation; prefer using `list[BaseEndpoint]`. https://peps.python.org/pep-0585/

resource_server_async/views/core.py
bcote-anl: I did not realize that. Thanks for letting me know! I changed all "Dict" to "dict"

resource_server_async/services.py
bcote-anl: Thanks for spotting! I changed all "List" to "list".

lowpolyneko: Missed a spot! 😉

resource_server_async/services.py
```diff
@@ -61,6 +61,26 @@
...
+    """Generate and return all endpoint adaptors for a given cluster."""
+
+    # For all endpoints in the database for this cluster ...
+    endpoint_adaptors: List[BaseEndpoint] = []
```
lowpolyneko: and here too :p

lowpolyneko (approved):

resource_server_async/services.py
bcote-anl: Totally

resource_server_async/services.py
bcote-anl: Indeed :) Thanks!
I fixed both in a new commit

PR #254, opened 2026-06-09
Uep status examples

bcote-anl:
Example (starting point) to investigate how to recover model status when using templated Globus Compute endpoints.


lowpolyneko (approved):
LGTM

PR #255, opened 2026-06-10
LCF-125 Batch timeout

bcote-anl:
A user had an issue with a batch stuck in "pending" for weeks. This is a task that is either lost in the Globus land, or lost due to a node failure on our cluster. This prevented the user from submitting a new batch because of the imposed quota of 2 active/pending batch per user. This PR adds an adjustable timeout to batches to fail them if it takes too long. It won't solve the whole problem, but at least it will bring some protection.


lowpolyneko (approved):
LGTM. Thanks for catching this!

PR #262, opened 2026-06-29
Add Minerva Direct API support

nscottnichols:
This adds Minerva as a Direct API-backed cluster. Direct API requests can now use optional mTLS config, Minerva model availability is read from the login-node status endpoint, and Minerva endpoints reject requests when a model is not live.

The direct health monitor now checks live Minerva models through their configured mTLS routes.

Validation:

- `python -m compileall resource_server_async cron_jobs`
- Minerva `/status` tested through mTLS from the login-node prototype


masalim2 (approved):
Looks great to me!

PR #263, opened 2026-06-30
LCF-131 Model alias

bcote-anl:
Suggested quick fix for the upcoming model name changes on Metis
e.g. gpt-oss-120b --> openai/gpt-oss-120b

The goal is to officially offer the new names (which will be consistent with the model names on Sophia), but still serve legacy names for a little while to avoid breaking existing user flows.


lowpolyneko (approved):
LGTM

PR #264, opened 2026-06-30
Storage alert

bcote-anl:
Now adding "df -h" command to alert if our storage rises above a certain percentage (defaulted to 80%).


lowpolyneko (approved):
LGTM, thanks for the quick feature :)

PR #265, opened 2026-07-01
LCF-106: Integrate AmSC D3 Triton Models

masalim2:
Integrates the gateway endpoints for the AmSC D3 HEP models. 

See https://github.com/masalim2/amsc-d3-triton/tree/main


bcote-anl (approved):
LGTM Thanks!

PR #266, opened 2026-07-13
schemas: allow extra fields in `extra_body` and `tools`

lowpolyneko:
This PR allows the bare minimum of extra fields to the OpenAI Chat Completion schema validation to allow Codex to work with FIRST.


masalim2 (approved):
Looks a pretty focused and conservative change, which is great. Thanks!

PR #268, opened 2026-07-23
Firstv2 benoit dev

bcote-anl:
### Adding skeletons for OpenAI API routes

This PR includes
- Pydantic classes for chat/completions, embeddings, responses (very simple with allow_extra)
- FastAPI dependencies to extract model from RouterConfig and verify permissions
- Added empty API routes for chat/completions, embeddings, responses

I tested with a dummy RouterConfig dictionary, and I was able to get a proper error when, not authenticated, authenticated but not in the group/domain, and when the model did not exist.

Next following this PR
- add non-federated deployment routes 
- add logic to build the AdmissionController from the resolved model object (including sorting deployments).


packages/common/first_common/schema/endpoints/openai.py
```diff
@@ -0,0 +1,56 @@
+from enum import Enum
+from typing import Any, List
```
masalim2:
```suggestion
from typing import Any
```

Prefer `list[T]` to `List[T]`

packages/common/first_common/schema/endpoints/openai.py
```diff
@@ -0,0 +1,56 @@
...
+    type: str | None = None
+    role: str | None = None
+    content: Any | None = None
+
+
+# https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create
+class OpenAIChatCompletionsPayload(BasePayload):
+    """OpenAI chat/completions spec."""
+
+    messages: list[ChatCompletionsMessage]
```
masalim2: I'm wondering if we will need any of these:

- ChatCompletionsMessage
- ResponsesInputItem

Is there any time when our gateway would introspect or use the content of `messages` or `input`?  If we do a true payload passthrough, maybe the only fields that are needed on our end are `model` and `stream`?

packages/common/first_common/schema/endpoints/openai.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+# https://platform.openai.com/docs/api-reference/embeddings/create
+class OpenAIEmbeddingsPayload(BasePayload):
+    """OpenAI embeddings spec."""
+
+    input: str | List[str] | List[int] | List[List[int]]
+
+
+class OpenAIEndpoints(Enum):
```
masalim2:
```suggestion
class OpenAIEndpoints(str, Enum):
```

Not 100% necessary but can be helpful because it allows `"chat/completions" == OpenAIEndpoints.chat_completions`

packages/common/first_common/schema/endpoints/openai.py
```diff
@@ -0,0 +1,56 @@
...
+    chat_completions = "chat/completions"
+    responses = "response"
```
masalim2:
```suggestion
    responses = "responses"
```

packages/gateway/first_gateway/apiserver/routes/openai/dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+
+from ....database.redis.router_config import ModelConfig
+from ...auth import enforce_permission
+from ...dependencies import AuthUser, RouterConfigDep
+
+
+async def resolve_model(
+    *, router_config: RouterConfigDep, user: AuthUser, model_name: str, endpoint: str
+) -> ModelConfig:
+    model = router_config.models_by_name.get(model_name)
```
masalim2: There's also a `router_config.models_by_alias` property we could check as a fallback here, just in case we want to continue using configuring aliases for backwards compatibility like `gpt-oss-120b -> openai/gpt-oss-120b`

packages/gateway/first_gateway/apiserver/routes/openai/dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+
+ResponsesModel = Annotated[
+    ModelConfig,
+    Depends(model_dependency(OpenAIEndpoints.responses.value)),
+]
+
+EmbeddingsModel = Annotated[
+    ModelConfig,
+    Depends(model_dependency(OpenAIEndpoints.embeddings.value)),
+]
```
masalim2: I think as written your dependencies look totally correct, but we could simplify a bit and make it more easily reusable.  If we removed the `model_dependency` wrapper and collapsed down to a single unified dependency: 

```python
async def resolve_model(
    request: Request,
    router_config: RouterConfigDep,
    user: AuthUser,
    payload: BasePayload,
) -> ModelConfig:
    model = router_config.models_by_name.get(payload.model)
    if model is None:
        raise NotFound(f"Model {payload.model} does not exist.")

    enforce_permission(user, model)

    endpoint = request.url.path.partition("/v1/")[2]
    if endpoint not in model.supported_endpoints:
        raise InvalidSpecError(f"Endpoint {endpoint} not supported for {model.name}.")
    return model

AuthorizedModel = Annotated[ModelConfig, Depends(resolve_model)]
```

That way, your `Depends(resolve_model)` is directly useable in any request, and we could re-use it for all endpoints and even for non-LLM dialects.

masalim2 (commented):
This looks fantastic, I'm so excited to see this half of it start coming alive!  The use of a single dependency to check model exists / authorized / endpoint supported is really clean!  

Just a few further questions/suggestions for simplifying the code.  Everything looks correct and it looks like now we have a great pattern in place for the API to use.

packages/common/first_common/schema/endpoints/openai.py
bcote-anl: Done

packages/common/first_common/schema/endpoints/openai.py
bcote-anl: Sounds good! I decided to keep "message" and "input" as mandatory field, but now it's just using Any without further pydantic models.

packages/common/first_common/schema/endpoints/openai.py
bcote-anl: That's nice indeed. 
Done.

packages/common/first_common/schema/endpoints/openai.py
bcote-anl: Good catch!
Done.

packages/gateway/first_gateway/apiserver/routes/openai/dependencies.py
bcote-anl: Great plan. I added the check after the first if statement.

packages/common/first_common/schema/endpoints/openai.py
bcote-anl: Sounds good! I'm now using Any instead of pydantic models to keep it light. I decided to keep the mandatory "messages" and "input" fields, but they are not restrictive.

packages/gateway/first_gateway/apiserver/routes/openai/dependencies.py
bcote-anl: Yes, that's much better, thanks! I applied the change and now the API routes can reuse the generic AuthorizedModel dependency.

masalim2 (approved):

PR #271, opened 2026-07-27
incorporating-admission-controler

bcote-anl:
### Incorporating AdmissionControler to OpenAI routes

This PR includes:
- new FIRST error classes
- AdmissionControler dependency
- service/orchestration.py file to extract target backend given model deployments
- submit_inference.py placeholder
- routing to specific deployment as well as federated routing


packages/common/first_common/errors.py
```diff
@@ -55,6 +55,21 @@ class NotImplemented(FirstError):
...
+
+
+class ServiceUnavailable(FirstError):
+    status_code = HTTPStatus.SERVICE_UNAVAILABLE
+    code: str = "service_unavailable"
+
+
+class InternalServerError(FirstError):
+    status_code = HTTPStatus.INTERNAL_SERVER_ERROR
+    code: str = "internal_server_error"
```
masalim2: We could consider resorting to the parent class `FirstError`, which also defaults to a 500 status code, as the least-specific "catch all" error type.  It's totally up to you though, this is fine too!

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
```diff
@@ -1,12 +1,59 @@
...
+    OpenAIEmbeddingsPayload,
+    OpenAIResponsesPayload,
+)
+
+from ....services.orchestration import get_backend
+from ....services.submit_inference import submit_inference
+from ...dependencies import AdmissionControllerDep, AuthUser
+from .dependencies import AuthorizedModel
+
+router = APIRouter(prefix="/deployments/{deployment_name}/v1")
```
masalim2: `schema/resources/plan_apply.py :: ResourceManifest.name`  allows each declared resource to contain forward slashes in the name.  My thinking was to allow deployment names like:

```yaml
kind: StaticDeployment
name: metis/google/gemma-4-31B-it # postgres UNIQUE constraint; stable identifier 
```

However, this prevents us from reliably using `name` in FastAPI Path parameters: if the deployment name contains "/", it won't parse correctly in the middle of a URL pattern like `/deployments/{name}/v1/messages`. 

My solution was to define the **slug** for each resource as `name.replace("/", "~")` (this is a property on every readable subclass of `ResourceMeta`).  The name regex was deliberately chosen to omit `~`, which is a URL-path-safe character, and therefore it gives us a 1:1 unique mapping between `name <---> slug`:

```python
name = slug.replace("~", "/")
slug = name.replace("/", "~")
```

Every other allowed character in the name is URL safe, and the mapping means we can reliably slugify/unslugify every resource name as needed.   

My suggestion would just be to change `{deployment_name}` --> `{deployment_slug}` here for URL path safety and use the slug translation. 

Other possibilities:

- disallow `/` in `name` ... would be simpler but slightly less pretty / more restrictive YAML config naming conventions
- You could update the `DeploymentConfig` schema to also include the slug, so the slug matching logic stays on the schema

I'm definitely open to your opinions here, this was just an arbitrary initial choice after deliberating about it in my own head!

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
```diff
@@ -1,12 +1,59 @@
...
+
+
+@router.post("/chat/completions", response_model=None)
+async def chat_completions(
+    deployment_name: str,
+    user: AuthUser,
+    payload: OpenAIChatCompletionsPayload,
+    model: AuthorizedModel,
+    admission_controler: AdmissionControllerDep,
```
masalim2:
```suggestion
    admission_controller: AdmissionControllerDep,
```

packages/gateway/first_gateway/apiserver/routes/openai/federated.py
```diff
@@ -9,28 +9,42 @@
...
 router = APIRouter(prefix="/federated/v1")
 
 
 @router.post("/chat/completions", response_model=None)
 async def chat_completions(
-    user: AuthUser, payload: OpenAIChatCompletionsPayload, model: AuthorizedModel
+    user: AuthUser,
+    payload: OpenAIChatCompletionsPayload,
+    model: AuthorizedModel,
+    admission_controler: AdmissionControllerDep,
```
masalim2:
```suggestion
    admission_controller: AdmissionControllerDep,
```

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
```diff
@@ -1,12 +1,59 @@
...
+) -> StreamingResponse | dict[str, Any]:
+    backend = await get_backend(
```
masalim2: This looks great.  I'm really a fan of the thin view and decoupling into 2 steps `get_backend` -> `submit_inference`.  It'll be easier to test these steps independently and the small view means that we can implement more inference endpoints with minimal code duplication!

Some other Admission Control considerations, not necessarily within the scope of this PR, but to handle eventually:

- **Generating a unique Request ID:**  `apiserver/api.py` registers the `log_request` middleware to run on every single HTTP request, which generates an `AccessLog` with unique UUID.  I copied this directly from FIRSTv1 as a starting point, but this could really benefit from your redesign.
    - What if we use the same UUID for both access log and request log?
    - What are the events that we want to emit structured logs for?
    - Can consider moving some logic from middleware into the `Depends()` system.  Then, for example, we could have dependencies for logging Token Usage and getting request IDs, without forcing that code pathway onto requests that don't actually prompt an LLM.  (The [dependencies with yield](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-with-yield/) pattern could be helpful for post-request actions like logging).
- **Settle:** the view should always call `admission_controller.settle()` in a `finally:` block when the request is over.  This ensures that each reservation held by admit is cleared when the request is over.  This frees up the Redis counters tracking backend concurrency / user concurrency and also fairly settles the user's token quota based on their actual usage.
- **Record error:** if a backend request fails, the view should also call `admission_controller.record_error()` to update the error counters in Redis.  This ensures that an unhealthy backend that's failing often gets benched for a timeout window.
- **Renewing Reservations:** A set of _currently active_ request IDs should be maintained in each api worker and passed into `admission_controller.renew()` every ~10 seconds (assuming `AdmissionController(lease_sec=30)`.  The 30 second expiration time means an apiserver that was holding 50 reservations and crashed will not keep Redis stuck in a corrupted "fully occupied" state forever.  At the same time, the heartbeating background asyncio task ensures that long/streaming reservations are properly held until `settle()` is called at the end.
- **Internal error retrying:** we could just try one backend and return the error to the user if the request fails, but a powerful option (what LiteLLM does) is to `record_error()` in Redis and transparently try the next backend in case of a failure.  Then, the user sees increased latency but they are shielded from the error, since our gateway handles the Retry.  

```python
# Pseudocode:

candidates = get_backend_candidates()
ac = admissionController

for attempt in range(3):
     backend = ac.admit(candidates)
     try:
         response = submit_inference(request_id, backend, payload)
     except BackendError:
         ac.record_error(backend) # count error for cooldown
         candidates.pop(backend) # don't try this one again
         token_usage = 0
         warning("backend failed; trying next one")
     else:
         token_usage = parse_token_usage(response) 
         return response
     finally: 
          ac.settle(request_id, actual_tokens=token_usage)  # every admit() is settle()'d
```

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -0,0 +1,25 @@
+from typing import Any, TypeVar
+
+from fastapi.responses import StreamingResponse
+from pydantic import BaseModel
+
+from ..database.redis.router_config import BackendConfig
+
+PayloadT = TypeVar("PayloadT", bound=BaseModel)
```
masalim2: Right now `BasePayload` is scoped within `schema.endpoints.openai`, but if we hoist this definition up to a higher level as the base class for _all_ inference payloads, then we can use that `BasePayload` as the type annotation in `submit_inference`.   

That'll help keep the type checking honest, and we will have a single subclass to extend for all inference payload types.

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -0,0 +1,25 @@
...
+async def submit_inference(
+    backend: BackendConfig,
+    payload: PayloadT,
+) -> StreamingResponse | dict[str, Any]:
+    """POST to an inference backend."""
+
+    # TODO: Submit and handle streaming / non-streaming
+    return {"Mock response": True}
+
+    payload = payload.model_dump(exclude_none=True)
```
masalim2:
```suggestion
    payload = payload.model_dump(exclude_unset=True, mode="json")
```

- If the user/agent explicitly sends a field set to `None` we can be faithful and allow that to pass through with `exclude_none=False` (default).
- At the same time, any field that they did not set in their JSON request and was populated with Pydantic defaults can be excluded with `exclude_unset=True`.  I think that keeps us as close as possible to retaining the user's original request/intent.   `mode="json"` ensures any weird python types like `datetime` are coerced to a value that won't error in `json.dumps()`

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -0,0 +1,25 @@
...
+from fastapi.responses import StreamingResponse
+from pydantic import BaseModel
+
+from ..database.redis.router_config import BackendConfig
+
+PayloadT = TypeVar("PayloadT", bound=BaseModel)
+
+
+async def submit_inference(
+    backend: BackendConfig,
```
masalim2, line 78: This can be for a future PR, but we'll probably want to expand our `BackendConfig` to track whether the client needs mTLS or not.  It'll probably be useful for every api worker to initialize some sort of class for tracking the correct `httpx.AsyncClient` instance per backend ID.  (I dunno if this makes sense to marry closely to the `RouterConfig` discovery worker, or have a separate class responsible for building clients when they're needed).  

When configured appropriately, the persistent, connection-pooling client objects per backend will be quite helpful to reduce the time and CPU overhead of establishing TLS / mTLS connections to each backend.

packages/gateway/first_gateway/services/orchestration.py
```diff
@@ -0,0 +1,163 @@
...
+
+    backend_id = await get_backend_id(user, model, admission_controler, deployment_name)
+    return next(
+        (
+            backend
+            for d in model.deployments
+            for backend in d.backends
+            if backend.id == backend_id
+        ),
+        None,
```
masalim2:
```suggestion
```

Can remove `None` fallback since this will raise if there is no backend

packages/gateway/first_gateway/services/orchestration.py
```diff
@@ -0,0 +1,163 @@
...
+            raise NotFound(f"Deployment {deployment_name} does not exist.")
+        deployments = [d]
+    else:
+        deployments = model.deployments
+
+    backend_candidates = get_candidates_from_deployments(deployments)
+
+    usage_limits = get_usage_limits(user, model.usage_limits)
+
+    # TODO: find where the request id is
```
masalim2: Yes, up to you!  Can use the log middleware that sets a global ContextVar, or create a new dependency for generating request IDs and inject the request_id here.  Middleware seems like it could be too heavy handed because there will be several API views that don't actually send an inference request.

packages/gateway/first_gateway/services/orchestration.py
```diff
@@ -0,0 +1,163 @@
...
+    # TODO: calculate estimated_tokens
```
masalim2: For token estimation, I think we could do surprisingly well with a dumb fast heuristic like ~4 characters per token `len(json.dumps(payload)) // 4 + max_output_tokens`.  It just needs to be a crude estimate for reservations, since it'll be corrected post-request by `settle()`.

- Estimate too low: the only negative consequence is that a user could sneak in additional prompts before hitting the quota
- Estimate too high: the negative consequence is that we will possibly quota-reject users too soon

OpenAI has released [tiktoken](https://github.com/openai/tiktoken) which is a fast Rust implementation of their tokenizer, so you can count tokens effectively before sending any prompt out.  It is still CPU intensive for large prompts and will consume quite a bit of CPU cores for a ~30% estimate improvement, so I don't know if it's worth it.

packages/gateway/first_gateway/services/orchestration.py
```diff
@@ -0,0 +1,163 @@
...
+            uid=backends[i].id,
+            max_backend_concurrency=router_params[i].max_backend_concurrency,
+            cooldown_threshold=router_params[i].cooldown_threshold,
+        )
+        for i in idx
+    ]
+
+
+def get_usage_limits(user: AuthUser, usage_policy: UsagePolicy) -> UsageLimits:
+    # TODO: check groups too, but what happen if more than 1 user groups are in overrides?
```
masalim2, line 118: Good question, this was just a rough initial sketch to support limits with easy overrides for testers and special users. Totally open to ideas to mature this design

packages/gateway/first_gateway/services/orchestration.py
```diff
@@ -0,0 +1,163 @@
...
+def raise_admit_error(admit_result: AdmitResult, usage_limits: UsageLimits) -> None:
+    if admit_result.status == AdmitStatus.REJECT_QUOTA:
+        if admit_result.quota_reason == QuotaReason.USER_CONCURRENCY:
+            raise TooManyRequests(
+                f"Concurent requests above f{usage_limits.max_user_concurrency}."
+            )
+        elif admit_result.quota_reason == QuotaReason.USER_RPM:
+            raise TooManyRequests(f"Requests per minute above f{usage_limits.rpm}.")
+        elif admit_result.quota_reason == QuotaReason.USER_TPM:
+            raise TooManyRequests(f"Tokens per minute above f{usage_limits.tpm}.")
```
masalim2: The `AdmitResult.retry_after_sec` should be set as an HTTP Response header `Retry-After: N` whenever it's available.  The Quota limits in particular will calculate the time remaining until the user can truly send the request again, so it will be quite useful to propagate that number back to the client

packages/gateway/first_gateway/services/orchestration.py
```diff
@@ -0,0 +1,163 @@
...
+        else:
+            raise InternalServerError(
+                f"Uncaught reject reason for status {AdmitStatus.REJECT_QUOTA}: {admit_result.quota_reason}."
+            )
+
+    elif admit_result.status == AdmitStatus.REJECT_CAPACITY:
+        if admit_result.capacity_reason == CapacityReason.SATURATED:
```
masalim2, line 164: In this case where we're at capacity, I think the `retry_after_sec` will be `None` but we can use the model `OverloadPolicy` as the configuration for how long to wait when the model is overloaded

masalim2 (commented):
This is fantastic, thank you for building this out!   Many of these comments are more forward-looking considerations for future PRs; nothing is blocking the merge of this step!  Let me know what you think!

packages/common/first_common/errors.py
bcote-anl: I like your plan better. I made the modification to use FirstError.

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
bcote-anl: I totally see your point! Let's go with your suggested conversion here. I added these two functions in the orchestration.py file (although maybe they should be put in some utils.py file somewhere?):

`def get_name_from_slug(slug: str) -> str:
    return slug.replace("~", "/")

def get_slug_from_name(name: str) -> str:
    return name.replace("/", "~")`
    
I now expect deployment_slug in the URL, which I then convert to name.

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
bcote-anl: Ah yes, of course! Fixed. Thanks!

packages/gateway/first_gateway/apiserver/routes/openai/federated.py
bcote-anl: Fixed

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
bcote-anl: Hi Misha, thanks for all these notes! While I agree with you that some of the feedback should be addressed in future PR to keep this one a bit more self-contained, I decided to incorporate your "attempt" logic since it was quite relevant to the current implementation. Now we have the following submit_inference_with_retry function that is called directly from the federated and deployment views:

```
async def submit_inference_with_retry(
    user: AuthUser,
    model: ModelConfig,
    admission_controller: AdmissionController,
    payload: PayloadT,
    deployment_name: str | None = None,
) -> StreamingResponse | dict[str, Any]:

    backend_candidates = get_backend_candidates(model, deployment_name=deployment_name)

    for attempt in range(len(backend_candidates)):
        backend = await get_backend(
            user,
            model,
            admission_controller,
            backend_candidates,
        )

        try:
            response = await submit_inference(backend, payload)
        except:
            deployment = get_deployment_from_backend(model.deployments, backend)
            admission_controller.record_error(backend.id, deployment.router_params)
            backend_candidates = [b for b in backend_candidates if b.uid != backend.id]
            token_usage = 0
            logger.warning(f"Backend {backend.id} failed, trying next one.")
        else:
            # TODO: implement parse_token_usage
            # token_usage = parse_token_usage(response)
            token_usage = 1  # TEMPORARY
            return response
        finally:
            # TODO: Incorporate request ID
            admission_controller.settle(
                "temporary_request_id", actual_tokens=token_usage
            )
```

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: Great plan! It's done. I moved `BasePayload` to `schema.endpoints.base` and we are now using it instead of `PayloadT`

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: Done.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: That is a great thing to consider. Let's indeed postpone to another PR. Should we raise an GitHub issue to not forget your thoughts?

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: Yes, I like it. I removed None from all my next() searches in this orchestration.py file

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: Thanks for the ideas! I will start a new PR for it and take into account your thoughts :)

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: I like the plan of keeping it a bit simple to not overuse the CPUs! Let's figure it out in another PR, but for now I've implemented `len(json.dumps(payload)) // 2 ` as a place holder, which should be similar to having a response similar in size then the input. That was my way to avoid figuring out max_output_tokens for now. But I left a TODO in the code to improve :)

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: I like the overwriting capability, but I am also unsure about what's the best way to handle it. But at least for now we have an enclosed function to expand on when we get to it. Thanks for all the ideas and thoughts so far!

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: Sounds good. I added retry_after_sec in the FirstError class (hopefully that's sufficient) and I'm now passing the value when raising the TooManyRequests for Quota limits

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: I like it! I now have a general block of code that will extract the appropriate value for the retry_after_sec:
```    
if admit_result.retry_after_sec:
        retry_after_sec = admit_result.retry_after_sec
        retry_str = f" Retry in {retry_after_sec} seconds."
    elif model.overload.short_retry_sec:
        retry_after_sec = model.overload.short_retry_sec
        retry_str = f" Retry in {retry_after_sec} seconds."
    else:
        retry_after_sec = None
        retry_str = ""
```
All of the rejection errors now append `retry_str` to their error message, and passes `retry_after_sec` to their raised exceptions

packages/common/first_common/errors.py
```diff
@@ -38,10 +38,12 @@ def __init__(
         *args: Any,
         status_code: HTTPStatus | int | None = None,
         info: dict[str, Any] | None = None,
+        retry_after_sec: int | None = None,
```
masalim2, line 27: This looks good! Now I'm realizing this should be exposed as a repsonse header in the exception handler of `api.py`, and that module has a separate handler for `TaskPending` just for the special case of `retry_after_sec`.

I like your approach better of having the retry header be optional on _any_ `FirstError`, so perhaps `TaskPending` can become another child class of `FirstError` and we can combine their exception handlers to set the header when retry time is not none.

packages/gateway/first_gateway/services/orchestration.py
masalim2: Looks perfect

packages/gateway/first_gateway/services/submit_inference.py
masalim2: yes, will do!

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -0,0 +1,84 @@
...
+    deployment_name: str | None = None,
+) -> StreamingResponse | dict[str, Any]:
+    """
+    Send inference requests to targeted backends using the admission
+    controler to select the appropriate backends. This function tries
+    to submit the request and return the response if successful, but
+    fallover to the next most appropriate backends in case of failures.
+    """
+
+    backend_candidates = get_backend_candidates(model, deployment_name=deployment_name)
```
masalim2, line 35: I think it's possible this returns 0 backends if all the deployments have zero weight?  Should we raise another `ServiceUnavailable` error in case there's no backends to the user?

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # TODO: figure out a better token estimate and max_output_tokens
+    # estimated_tokens = len(json.dumps(payload)) // 4 + max_output_tokens
+    estimated_tokens = len(json.dumps(payload)) // 2
+
+    for attempt in range(len(backend_candidates)):
```
masalim2:
```suggestion
    for attempt in range(min(3, len(backend_candidates))):
```

We could consider bounding this in case there's a lot of backends that are all failing for some reason.

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -0,0 +1,84 @@
...
+        else:
+            # TODO: implement parse_token_usage
+            # token_usage = parse_token_usage(response)
+            token_usage = 1  # TEMPORARY
+            return response
+        finally:
+            # TODO: Incorporate request ID
+            admission_controller.settle(
+                "temporary_request_id", actual_tokens=token_usage
+            )
```
masalim2, line 71: At the end, if we fall out of the retry loop without success on any backend, we should raise an error.  Perhaps we can bind the encountered error (`except exc`) so we can re-raise it if nothing worked at the end.

packages/gateway/first_gateway/services/orchestration.py
masalim2: One last thing is to wire it through the exception handler for FirstError in api.py to set the HTTP Response header.  Mentioned in another comment how this now could be be merged with `TaskPending`

masalim2:
Looking great! Just a handful of minor comments to address on this PR and I will populate some issues to round up the remaining ideas that we discussed for future PRs!

packages/gateway/first_gateway/services/orchestration.py
bcote-anl: Could you have a look at errors.py and error_handlers.py? I think I made the appropriate changes. TaskPending now inherits from FirstError, and in the handler it looks for the retry_in_sec variable before populating the headers. I haven't spent much time on the error handling side for V2, so my changes may be a bit simplistic?

packages/common/first_common/errors.py
bcote-anl: I incorporated this comment in the change above.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: That's a good idea! I added the if statement. I did not raise the error in the get_backend_candidates function directly in case we want to use this function elsewhere internally for tests.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: I like it. Done.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: I added a raise ServiceUnavailable error if we get our of the "attempt" loop. I decided, however, to not bind the encountered error just yet, since it might require some extra coding and refactoring. I think it would be best to not send the full tracebacks to user, so instead we might want to more carefully parse and display what went wrong with the backends. I'm wondering if we should provide error ID for backend failure (just like we do with uncaught errors) or elaborate on what type of exceptions we get from submit_inference. I like the idea, I'm just not sure yet what's the optimal way to add this feature... Maybe we can raise an issue for it?

packages/gateway/first_gateway/services/orchestration.py
masalim2: It looks spot on, thank you!

packages/gateway/first_gateway/services/submit_inference.py
masalim2: Looks good!  I'll add an issue for the future!

masalim2: @bcote-anl added #302 for future

PR #277, opened 2026-08-05
Add Minerva session affinity and cache isolation

nscottnichols:
## Summary

Adds cache-aware routing for Minerva so related requests stay on the same vLLM replica while prefix-cache data remains isolated between users.

## Changes

- Accepts an optional X-ALCF-Session-ID.
- Derives opaque, server-controlled affinity keys from the authenticated user, model, and session.
- Adds per-user/model cache_salt values for supported vLLM operations.
- Prevents callers from overriding internal affinity or cache-isolation values.
- Keeps streaming and concurrent requests isolated.
- Applies only to Minerva adapters; other backends are unchanged.
- Documents the required MINERVA_AFFINITY_HMAC_KEY secret.

  ## Validation

- Full FIRST suite: 493 tests passed.
- Deployed and tested on the dev VM.
- Verified non-streaming, streaming, concurrency, user isolation, and that sensitive values do not appear in logs.

### DIFF
--- a/inference_gateway/settings.py
+++ b/inference_gateway/settings.py
@@ -37,6 +37,10 @@ load_dotenv(override=True)
 # Django secret key
 SECRET_KEY = os.getenv("SECRET_KEY")
 
+# Dedicated Minerva affinity/cache namespace key. It is validated only when a
+# Minerva inference request is forwarded so unrelated adapters remain usable.
+MINERVA_AFFINITY_HMAC_KEY = os.getenv("MINERVA_AFFINITY_HMAC_KEY")
+
 # Define whether we are in the automated test suite mode
 RUNNING_AUTOMATED_TEST_SUITE = os.getenv(
     "RUNNING_AUTOMATED_TEST_SUITE", "False"
--- a/resource_server_async/endpoints/direct_api.py
+++ b/resource_server_async/endpoints/direct_api.py
@@ -1,4 +1,5 @@
 import asyncio
+import copy
 import json
 import logging
 import os
@@ -97,13 +98,24 @@ class DirectAPIEndpoint(BaseEndpoint):
 
     # Submit task
     async def submit_task(self, data: dict[str, Any]) -> SubmitTaskResult:
+        return await self._submit_task_with_headers(data)
+
+    async def _submit_task_with_headers(
+        self,
+        data: dict[str, Any],
+        request_headers: dict[str, str] | None = None,
+    ) -> SubmitTaskResult:
         """Submits a single interactive task to the compute resource."""
-        endpoint = data.pop("openai_endpoint", "chat/completions").strip("/")
+        request_data = copy.deepcopy(data)
+        endpoint = request_data.pop("openai_endpoint", "chat/completions").strip("/")
         url = f"{self.config.api_url.rstrip('/')}/{endpoint}"
+        captured_headers = dict(request_headers) if request_headers else None
 
         # Submit POST call and wait for the response
         try:
-            response = await self.httpx_client.post(url, data=data)
+            response = await self.httpx_client.post(
+                url, data=request_data, headers=captured_headers
+            )
         except httpx.HTTPStatusError as e:
             raise EndpointError(
                 f"Upstream endpoint returned {e.response.status_code}: {e.response.content[:256]!r}.",
@@ -126,7 +138,24 @@ class DirectAPIEndpoint(BaseEndpoint):
     async def submit_streaming_task(
         self, data: dict[str, Any]
     ) -> SubmitStreamingTaskResponse:
+        return await self._submit_streaming_task_with_headers(data)
+
+    async def _submit_streaming_task_with_headers(
+        self,
+        data: dict[str, Any],
+        request_headers: dict[str, str] | None = None,
+    ) -> SubmitStreamingTaskResponse:
         """Submits a single interactive task to the compute resource with streaming enabled."""
+
+        # Capture request-local state before returning the lazy streaming
+        # generator. Endpoint adapters and HTTP clients are shared objects.
+        request_data = copy.deepcopy(data)
+        endpoint = request_data.pop("openai_endpoint", "chat/completions").strip("/")
+        url = f"{self.config.api_url.rstrip('/')}/{endpoint}"
+        captured_headers = {
+            **self.httpx_client.headers,
+            **dict(request_headers or {}),
+        }
 
         # Shared state for tracking streaming (optimized - minimal memory)
         streaming_state: StreamingState = {
@@ -143,7 +172,9 @@ class DirectAPIEndpoint(BaseEndpoint):
 
             # For each streaming chunk ...
             try:
-                async for chunk in self.__get_stream_chunks(data):
+                async for chunk in self.__get_stream_chunks(
+                    url, request_data, captured_headers
+                ):
                     if chunk:
                         # Send chunk
                         streaming_state["total_chunks"] += 1
@@ -205,12 +236,12 @@ class DirectAPIEndpoint(BaseEndpoint):
 
     # Get stream chunks
     async def __get_stream_chunks(
-        self, data: dict[str, Any]
+        self,
+        url: str,
+        data: dict[str, Any],
+        headers: dict[str, str],
     ) -> AsyncGenerator[str, None]:
         """Make a direct API streaming call to the endpoint."""
-        endpoint = data.pop("openai_endpoint", "chat/completions").strip("/")
-        url = f"{self.config.api_url.rstrip('/')}/{endpoint}"
-
         # Create an async HTTPx client
         try:
             async with httpx.AsyncClient(
@@ -228,7 +259,7 @@ class DirectAPIEndpoint(BaseEndpoint):
                     "POST",
                     url,
                     json=data,
-                    headers=self.httpx_client.headers,
+                    headers=headers,
                 ) as response:
                     # Return error if something went wrong
                     if response.status_code != 200:
--- a/resource_server_async/endpoints/direct_api.py
+++ b/resource_server_async/endpoints/direct_api.py
@@ -1,5 +1,4 @@
 import asyncio
-import copy
 import json
 import logging
 import os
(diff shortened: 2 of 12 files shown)


resource_server_async/endpoints/direct_api.py
```diff
@@ -97,13 +98,24 @@ def __init__(
 
     # Submit task
     async def submit_task(self, data: dict[str, Any]) -> SubmitTaskResult:
+        return await self._submit_task_with_headers(data)
```
lowpolyneko, line 128: No need for the functions to be split out IMO

resource_server_async/endpoints/direct_api.py
```diff
@@ -97,13 +98,24 @@ def __init__(
...
+
+    async def _submit_task_with_headers(
+        self,
+        data: dict[str, Any],
+        request_headers: dict[str, str] | None = None,
+    ) -> SubmitTaskResult:
         """Submits a single interactive task to the compute resource."""
-        endpoint = data.pop("openai_endpoint", "chat/completions").strip("/")
+        request_data = copy.deepcopy(data)
+        endpoint = request_data.pop("openai_endpoint", "chat/completions").strip("/")
```
lowpolyneko, line 137: IIRC this `.pop()` was in-place as a workaround for Metis. @masalim2 do you remember why this was needed?

resource_server_async/logging.py
```diff
@@ -26,6 +27,7 @@ class RequestContext:
     access_log: AccessLogPydantic
     user: UserPydantic | None = None
     request_log: RequestLogPydantic | None = None
+    minerva_session_id: str | None = None
```
lowpolyneko, line 30: In the future, we'll probably need to keep a session ID for other clusters for similar reasons. This is fine for now though.

lowpolyneko (changes requested):
Looks good apart from some minor nits. Thanks for the quick work :)

resource_server_async/minerva_affinity.py
```diff
@@ -0,0 +1,119 @@
...
+)
+
+
+class MinervaAffinityConfigurationError(RuntimeError):
+    pass
+
+
+def validate_session_id(value: str | None) -> str | None:
+    """Validate an optional visible-ASCII conversation/branch identifier."""
+    if value is None or value == "":
```
lowpolyneko, line 42:
```suggestion
    if not value:
```

resource_server_async/endpoints/direct_api.py
masalim2: `data` is the POST request body plus the key `openai_endpoint` that doesn't belong in the request body (it informs the URL path like `/v1/chat/completions`).  

So `pop()` is the correct logic to remove it from the POST payload before sending it upstream.  At best the LLM backend will ignore the unknown key; at worst, it causes a validation error.

inference_gateway/settings.py
```diff
@@ -37,6 +37,10 @@
 # Django secret key
 SECRET_KEY = os.getenv("SECRET_KEY")
 
+# Dedicated Minerva affinity/cache namespace key. It is validated only when a
+# Minerva inference request is forwarded so unrelated adapters remain usable.
+MINERVA_AFFINITY_HMAC_KEY = os.getenv("MINERVA_AFFINITY_HMAC_KEY")
```
masalim2, line 42: Rather than add this to persistent configuration, consider generating this semi-persistent data on the fly and storing it in Redis.  It should be just a few extra lines of logic and avoid the need for another secret to manage and rotate in the deployment.  

I'd advocate for generating a string with `secrets` and then atomically inserting it in Redis with `SET key value NX EX ttl`. As long as we're still using Django cache, the `add(key, value, timeout)` method wraps this redis `SET NX` behavior.  The `NX` flag makes it set key=value only if key doesn't already exist.

- API workers grab the key from redis when they need it. If missing, they atomically `cache.add()` it, which is safe against race conditions, with a ~24 hour TTL and re-get the added key.
- When the TTL expires it naturally rotates; a 24 hour window would make sessions crossing the daily prefix invalidation time pretty rare.

resource_server_async/endpoints/direct_api.py
```diff
@@ -97,13 +98,24 @@ def __init__(
...
+        return await self._submit_task_with_headers(data)
+
+    async def _submit_task_with_headers(
+        self,
+        data: dict[str, Any],
+        request_headers: dict[str, str] | None = None,
+    ) -> SubmitTaskResult:
         """Submits a single interactive task to the compute resource."""
-        endpoint = data.pop("openai_endpoint", "chat/completions").strip("/")
+        request_data = copy.deepcopy(data)
```
masalim2: What's the rationale for creating a copy here?  I think `data` is built for the request and discarded and safe to mutate in place.

resource_server_async/endpoints/direct_api.py
```diff
@@ -125,9 +137,26 @@ async def submit_task(self, data: dict[str, Any]) -> SubmitTaskResult:
...
         """Submits a single interactive task to the compute resource with streaming enabled."""
 
+        # Capture request-local state before returning the lazy streaming
+        # generator. Endpoint adapters and HTTP clients are shared objects.
+        request_data = copy.deepcopy(data)
+        endpoint = request_data.pop("openai_endpoint", "chat/completions").strip("/")
+        url = f"{self.config.api_url.rstrip('/')}/{endpoint}"
+        captured_headers = {
+            **self.httpx_client.headers,
+            **dict(request_headers or {}),
```
masalim2: I might reverse the order of unpacking here so that a client can't overwrite any headers that our `httpx_client` wants to have set. It may also be safer to limit the set of passed headers to a well-known set of `X-*` keys.  That immediately shoots down clients who try to get sneaky and inject untrusted headers (even though other parts of the stack should be hardened against that)

resource_server_async/minerva_affinity.py
```diff
@@ -0,0 +1,119 @@
...
+    return hmac_base64url(secret, ("minerva-cache-v1", user_id, model_slug))
+
+
+def derive_affinity_key(
+    secret: bytes,
+    user_id: str,
+    model_slug: str,
+    session_id: str | None,
+) -> str:
+    return hmac_base64url(
```
masalim2, line 90: Actually, since we're the ones deriving the affinity key and sending it upstream, I wonder if we even need to use HMAC with a secret in the first place.  Why not simplify and use SHA256 without a secret key?

resource_server_async/minerva_affinity.py
```diff
@@ -0,0 +1,119 @@
...
+        secret,
+        (
+            "minerva-affinity-v1",
+            user_id,
+            model_slug,
+            session_id if session_id is not None else "user-default",
```
masalim2, line 96: Does the affinity map on the Minerva side have a finite expiration time so that users aren't pinned to a specific replica by default forever?

resource_server_async/minerva_affinity.py
nscottnichols: I don't want falsy here. Only `None` and `""` are intended to be None

resource_server_async/minerva_affinity.py
nscottnichols: There was two reasons here. 1. I was "worried" about potential collisions in SHA256. 2. I figured we might as well test a more secure thing here for potential future CUI work before we need to build something out.

resource_server_async/minerva_affinity.py
nscottnichols: Yes, this is on the NGINX side.

inference_gateway/settings.py
nscottnichols: Great! I was actually unsure of how to best approach this so went with the easier thing hoping you would come up with something better. I'll see if I can make it work the way you describe.

resource_server_async/minerva_affinity.py
masalim2: I think we can treat SHA256 as collision safe and fully secure in this context.  HMAC adds an authentication layer on top, so that other parties that also hold the secret can verify the string was signed by another secretholder. 

I don't think we have that aspect in this system, we just need a way to turn a `(user, model, session)` into a hash blob that has 1. ~zero collision risk and 2. ~impossible to reverse-engineer and I think SHA256 checks both of those boxes!

inference_gateway/settings.py
masalim2: (irrelevant if we can use a secure hash without message authenticate code)

resource_server_async/endpoints/direct_api.py
nscottnichols: This is because when I was originally messing around, I was modifying the request data and I never went back to clean this up. Will revert the deep copies in this section.

resource_server_async/endpoints/direct_api.py
nscottnichols: Now that I think about this, we probably need more than just reversing the order here... let me see if I can come up with a minimal solution.

inference_gateway/settings.py
nscottnichols: @masalim2 Adding this is a bit more complex than I expected and I need to do more reading on it. Can we move forward with the current secret approach so I can get Minerva working with replicas and I can come back and make this change in a future PR?

masalim2: yes for sure!

resource_server_async/endpoints/direct_api.py
nscottnichols: Fixed in [cc6425e](https://github.com/argonne-lcf/FIRST/pull/277/commits/cc6425ede5227f7e02e6d4cbcdbd67b1d84e8270)

resource_server_async/endpoints/direct_api.py
nscottnichols: This should now be covered in [b9a64f6](https://github.com/argonne-lcf/FIRST/pull/277/commits/b9a64f6f17253c23b29f571969262e9b061d4785). I added an explicit allow list and did some casing. The `test_nonstream_rejects_mixed_case_default_header_overrides` and `test_stream_rejects_mixed_case_default_header_overrides` should cover reject paths for any sneaky headers, but let me know if you want an explicit test.

masalim2: LGTM!

PR #278, opened 2026-08-05
resource_server_async/models: drop unused models

lowpolyneko:
Since we've migrated to structured logging, we've reduced our reliance on Postgres for only batch requests, endpoints, and clusters. We can remove these models from the database now that it's dead code.


bcote-anl:
Looks like a good plan to keep the codes cleaner. Just want to make sure that BatchMetrics is indeed not used anymore since Batch still relies on Postgres?

bcote-anl (approved):

PR #279, opened 2026-08-07
Separate FIRST validation from backend request schemas

nscottnichols:
## Summary

Separates FIRST-owned request validation from backend API validation. FIRST now validates only fields it uses for routing, authorization, streaming, accounting, and internal adapter behavior while forwarding backend-specific JSON unchanged.

## Key behavior changes

- Unknown backend fields—including top-level chat_template_kwargs—pass through without requiring FIRST schema changes.
- Literal nested extra_body remains nested and unchanged.
- Backend-owned validation errors now come from the selected backend instead of FIRST.
- Top-level FIRST-reserved fields such as cache_salt, api_port, route metadata, and Globus streaming metadata are rejected.
- FIRST remains authoritative for model routing, stream mode, upstream authorization, affinity headers, and internal metadata.
- Direct API, Minerva, Metis, and Globus adapters preserve request data using shallow per-request copies.
- Exhaustive OpenAI request schemas and obsolete backend-validation fixtures were removed.
- Full request payloads are no longer written to debug logs (should I restore this behavior?).


lowpolyneko (approved):
Thanks for cleaning this up!

lowpolyneko:
> Full request payloads are no longer written to debug logs (should I restore this behavior?).

I think it's better this way. Invalid requests will still be logged in the `request_log` slog stream as it gets proxied to the backend inference endpoint.

PR #280, opened 2026-08-12
Add v2->v1 bridge

masalim2:
This PR adds a small bridge service to V1 (`main`) that reads the V2 controller's `RouterConfig` JSON blob and automates the create/update/delete of V1 `Endpoints`. 

- New `FirstV2Endpoint` Direct API endpoint class with mTLS passthrough and optional/temporary http proxying for development.  One endpoint's config holds multiple backend URLs and randomly chooses a backend URL at request time.
- New `TaraCluster` Cluster adapter class which reflects the RouterConfig into `get_jobs()` output
- bridge entrypoint shares the main Django app's dotenv settings and `manage.py` runner
- A systemd unit to run the bridge persistently
- V2->V1 Endpoint mapping with deterministic PKs derived from SHA1 of `(cluster, model)` and pass-through of allowed Globus groups and domains

### DIFF
--- a/resource_server_async/first_v2_bridge/mapping.py
+++ b/resource_server_async/first_v2_bridge/mapping.py
@@ -1,0 +1,131 @@
+"""Pure mapping functions from a parsed RouterConfig to desired V1 Endpoint rows."""
+
+import hashlib
+from dataclasses import dataclass, field
+from typing import Any
+
+from django.utils.text import slugify
+
+from .router_config import ModelConfig, RouterConfig
+from .settings import BridgeSettings
+
+ENDPOINT_ADAPTER = "resource_server_async.endpoints.first_v2.FirstV2Endpoint"
+
+# Bridge-managed rows use a deterministic high PK range so they never collide
+# with fixture-loaded admin rows.
+PK_BASE = 2**40
+_PK_SPAN = 2**40
+
+# V2 lists only healthy replicas, and only pilot deployments are proxied by the
+# bridge.
+_PILOT_KIND = "pilot"
+
+
+def deterministic_pk(cluster: str, model: str) -> int:
+    """Stable high PK keyed on the model identity (not backend id).
+
+    Keying on ``(cluster, model)`` means replica churn is an in-place UPDATE of
+    ``model_urls`` rather than a slug-thrashing delete+insert.
+    """
+    digest = hashlib.sha1(f"{cluster}/{model}".encode()).hexdigest()
+    return PK_BASE + int(digest, 16) % _PK_SPAN
+
+
+@dataclass
+class DesiredEndpoint:
+    pk: int
+    endpoint_slug: str
+    cluster: str
+    framework: str
+    model: str
+    endpoint_adapter: str
+    allowed_globus_groups: list[str]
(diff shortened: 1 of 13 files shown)


resource_server_async/management/commands/first_v2_bridge.py
```diff
@@ -0,0 +1,120 @@
...
+        self.stdout.write(
+            f"first_v2_bridge starting poll={settings.poll_interval_sec}s"
+        )
+
+        while True:
+            try:
+                self._tick(settings)
+            except Exception as e:
+                log.exception("first_v2_bridge tick failed: %s", e)
+            time.sleep(settings.poll_interval_sec)
```
nscottnichols, line 60: The tick interval will be inconsistent based on how long `_tick()` takes, but I say good enough!

resource_server_async/first_v2_bridge/router_config.py
```diff
@@ -0,0 +1,77 @@
...
+    models: list[ModelConfig] = []
+
+    @classmethod
+    def load(cls, client: "redis.Redis") -> "RouterConfig":
+        """Read and parse the router config blob from Redis (sync client)."""
+        raw = client.get(ROUTER_CONFIG_KEY)
+        if not raw:
+            return cls()
+        assert isinstance(raw, (str, bytes, bytearray))
+        return cls.model_validate_json(raw)
```
nscottnichols, line 77: The default return for `redis.Redis.get("missing-key")` is `None`. So we should probably add a no-op case for that. I'm not sure if this is a realistic concern though. I'm thinking if there is some interruption and we start the V2 controller and bridge in the wrong order, the entries could get dropped. But I guess they would be repopulated on the next tick.

resource_server_async/first_v2_bridge/mapping.py
```diff
@@ -0,0 +1,131 @@
...
+
+from .router_config import ModelConfig, RouterConfig
+from .settings import BridgeSettings
+
+ENDPOINT_ADAPTER = "resource_server_async.endpoints.first_v2.FirstV2Endpoint"
+
+# Bridge-managed rows use a deterministic high PK range so they never collide
+# with fixture-loaded admin rows.
+PK_BASE = 2**40
+_PK_SPAN = 2**40
```
nscottnichols, line 17: Is there risk of overflow here? What is the datatype for the PK's final destination?

nscottnichols (commented):
We need to ensure the Tara cluster fixture is there before running the bridge code.

resource_server_async/first_v2_bridge/mapping.py
masalim2: I did check that the PK column is a bigint with a max value of `2**63 - 1`, so we should be okay at this range.

resource_server_async/first_v2_bridge/router_config.py
masalim2: Good call; I've addressed this by returning `None` if the config key is missing from redis rather than populating defaults.  Then the tick will no-op and we won't clear out backends needlessly.

nscottnichols (approved):
LGTM!

PR #282, opened 2026-08-17
Make GraphQL PBS job state authoritative

nscottnichols:
## Summary

- Paginate active GraphQL PBS jobs and fail closed on malformed, duplicate, or unknown scheduler data.
- Add exact-job lookup so omission from a bulk listing is never treated as proof that an allocation is gone.
- Validate deletion requests and wait for the exact job to become terminal or absent.
- Publish scheduler-derived endpoints only for running jobs.

## Why this is needed

A paginated listing can omit a live allocation, while a successful delete mutation is only an acknowledgement. Treating either as proof of release could free capacity while the scheduler still owns the nodes.

## Tests

- `tests/test_graphql_pbs.py`: 8 passed

### DIFF
--- a/packages/common/first_common/schema/base_scheduler.py
+++ b/packages/common/first_common/schema/base_scheduler.py
@@ -141,8 +141,8 @@ class SchedulerAdapter(ABC):
         """
         Request termination of a job in the scheduler [qdel].
 
-        Returning means the scheduler accepted the request; callers must
-        observe the job as gone before treating its allocation as released.
+        Returning means the scheduler accepted the request. Scheduler cleanup
+        may continue asynchronously after the acknowledgement.
         """
 
     @abstractmethod
--- a/packages/common/first_common/schema/base_scheduler.py
+++ b/packages/common/first_common/schema/base_scheduler.py
@@ -117,6 +117,25 @@ class SchedulerAdapter(ABC):
         List job statuses from the scheduler [qstat]
         """
 
+    async def get_exact_job_status(self, job_id: str) -> JobStatusInfo | None:
+        """Return one exact job, but never infer absence from a bulk listing.
+
+        Adapters with an authoritative exact-ID lookup should override this and
+        may return ``None`` only when that lookup explicitly proves absence.
+        The conservative default can return an exact match from the ordinary
+        list, but fails closed if it is omitted (for example by pagination).
+        """
+        matches = [
+            status for status in await self.get_job_statuses() if status.id == job_id
+        ]
+        if len(matches) == 1:
+            return matches[0]
+        if len(matches) > 1:
+            raise RuntimeError(f"scheduler returned duplicate exact job ID {job_id!r}")
+        raise RuntimeError(
+            f"scheduler adapter cannot prove exact job {job_id!r} is absent"
+        )
+
     @abstractmethod
     async def terminate_job(self, job_id: str) -> None:
         """
--- a/packages/common/first_common/schema/base_scheduler.py
+++ b/packages/common/first_common/schema/base_scheduler.py
@@ -118,7 +118,7 @@ class SchedulerAdapter(ABC):
         """
 
     async def get_exact_job_status(self, job_id: str) -> JobStatusInfo | None:
-        """Return one job, returning None only if it is definitively absent. 
+        """Return one job, returning None only if it is definitively absent.
 
         Adapters with an authoritative exact-ID lookup should override this and
         may return ``None`` only when that lookup explicitly proves absence.
@@ -139,7 +139,10 @@ class SchedulerAdapter(ABC):
     @abstractmethod
     async def terminate_job(self, job_id: str) -> None:
         """
-        Terminate a job in the scheduler [qdel]
+        Request termination of a job in the scheduler [qdel].
+
+        Returning means the scheduler accepted the request; callers must
+        observe the job as gone before treating its allocation as released.
         """
 
     @abstractmethod
--- a/packages/common/first_common/schema/base_scheduler.py
+++ b/packages/common/first_common/schema/base_scheduler.py
@@ -118,12 +118,12 @@ class SchedulerAdapter(ABC):
         """
 
     async def get_exact_job_status(self, job_id: str) -> JobStatusInfo | None:
-        """Return one exact job, but never infer absence from a bulk listing.
+        """Return one job, returning None only if it is definitively absent. 
 
         Adapters with an authoritative exact-ID lookup should override this and
         may return ``None`` only when that lookup explicitly proves absence.
-        The conservative default can return an exact match from the ordinary
-        list, but fails closed if it is omitted (for example by pagination).
+        The default returns an exact match from the job listing
+        but fails closed if it is omitted (for example by pagination).
         """
         matches = [
             status for status in await self.get_job_statuses() if status.id == job_id
--- a/packages/common/first_common/schema/base_scheduler.py
+++ b/packages/common/first_common/schema/base_scheduler.py
@@ -142,7 +142,8 @@ class SchedulerAdapter(ABC):
         Request termination of a job in the scheduler [qdel].
 
         Returning means the scheduler accepted the request. Scheduler cleanup
-        may continue asynchronously after the acknowledgement.
+        may continue asynchronously after the acknowledgement; callers that
+        need allocation-release proof must observe the job as gone.
         """
 
     @abstractmethod
--- a/packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
+++ b/packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
@@ -271,11 +271,7 @@ class GraphQLPBSAdapter(SchedulerAdapter):
                     seen_ids.add(status.id)
                     results.append(status)
 
-            page_info = connection.get("pageInfo")
-            if not isinstance(page_info, dict) or not isinstance(
-                page_info.get("hasNextPage"), bool
-            ):
-                raise RuntimeError("GraphQL jobs query returned malformed pageInfo")
+            page_info = data["jobs"]["pageInfo"]
             if not page_info["hasNextPage"]:
                 return results
             next_cursor = page_info.get("endCursor")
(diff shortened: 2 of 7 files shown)


packages/common/first_common/schema/base_scheduler.py
```diff
@@ -117,6 +117,25 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
         List job statuses from the scheduler [qstat]
         """
 
+    async def get_exact_job_status(self, job_id: str) -> JobStatusInfo | None:
+        """Return one exact job, but never infer absence from a bulk listing.
+
+        Adapters with an authoritative exact-ID lookup should override this and
+        may return ``None`` only when that lookup explicitly proves absence.
+        The conservative default can return an exact match from the ordinary
+        list, but fails closed if it is omitted (for example by pagination).
```
masalim2:
```suggestion
        """Return one job, returning None only if it is definitively absent. 

        Adapters with an authoritative exact-ID lookup should override this and
        may return ``None`` only when that lookup explicitly proves absence.
        The default returns an exact match from the job listing
        but fails closed if it is omitted (for example by pagination).
```

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -74,6 +81,40 @@ def _head_node_hostname(machines: list[dict[str, Any]]) -> str | None:
...
+    head_ip = None
+    head_hostname = None
+    if state_code in _ACTIVE_STATES:
+        machines = node.get("allocatedMachines") or []
+        head_ip = _head_node_ip(machines)
+        head_hostname = _head_node_hostname(machines)
+
+    return JobStatusInfo(
+        id=job_id,
+        name=node.get("name") or "",
```
masalim2:
```suggestion
        name=node["name"],
```

Let's fail on missing name, since that ends up getting used by the controller in some other places. It should be okay to just let this fail loudly on the default `KeyError`.

lowpolyneko (approved):
LGTM, I'll defer to @masalim2 for the merge though.

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -193,71 +243,189 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
...
+                node = edge.get("node")
+                if not isinstance(node, dict):
+                    raise RuntimeError("GraphQL jobs edge returned no node")
+                status = _job_status_from_node(node)
+                if status.id in seen_ids:
+                    raise RuntimeError(
+                        f"GraphQL jobs pages duplicated job ID {status.id!r}"
+                    )
+                seen_ids.add(status.id)
+                results.append(status)
```
masalim2:
```suggestion
            edges: list[dict[str, Any]] = data["jobs"]["edges"]
            for edge in edges:
                if edge.get("error"):
                    raise RuntimeError(f"GraphQL jobs edge failed:\n{edge['error']}")
                node: dict[str, Any] = edge["node"]
                status = _job_status_from_node(node)
                if status.id not in seen_ids:
                    seen_ids.add(status.id)
                    results.append(status)
```

I think this block could be a little less defensive, as suggested above.  It's a great call to raise the `RuntimeError` on errors and fetch all pages in the listing, since the controller infers missing jobs are done.

However, I think we can safely dedup rather than throwing a `RuntimeError` on duplicates, and I think it's a bit easier to read if the type-checking `RuntimeErrors` are omitted as suggested.  It'll still fail closed on bad data from GraphQL with the native `AttributeError` or `KeyError`.  Since this Adapter is built-for-purpose I think those types of errors would be just as illuminating in a Traceback while keeping the intent of the code clear.

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -193,71 +243,189 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
...
+                return results
+            next_cursor = page_info.get("endCursor")
+            if (
+                not isinstance(next_cursor, str)
+                or not next_cursor
+                or next_cursor in seen_cursors
+            ):
+                raise RuntimeError("GraphQL jobs pagination cursor did not advance")
+            seen_cursors.add(next_cursor)
+            cursor = next_cursor
```
masalim2:
```suggestion
            page_info = data["jobs"]["pageInfo"]
            if not page_info["hasNextPage"]:
                return results
            next_cursor = page_info.get("endCursor")
            if (
                not isinstance(next_cursor, str)
                or not next_cursor
                or next_cursor in seen_cursors
            ):
                raise RuntimeError("GraphQL jobs pagination cursor did not advance")
            seen_cursors.add(next_cursor)
            cursor = next_cursor
```

Similarly here not sure if we need to be extra defensive on the correctness of the GraphQL's server's pagination but this is fine.

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -193,71 +243,189 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
...
+        data = await self._post(query, {"jobId": job_id})
         payload = data["deleteJob"]
         if payload.get("error"):
+            # A retry after a lost acknowledgement can legitimately receive a
+            # not-found mutation error.  The error alone is never release
+            # evidence; accept it only if a separate exact lookup proves the
+            # same ID absent or gone.
+            state = await self._get_exact_job_state(job_id)
+            if state is None or state == SchedulerJobState.gone:
+                return
```
masalim2, line 313:
```suggestion
            if state is None or state == SchedulerJobState.gone:
                logger.warning(f"terminate {job_id=} error: job is already gone.")
                return
```

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -193,71 +243,189 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
...
+        }
         """
-        data = await self._post(query)
+        data = await self._post(query, {"jobId": job_id})
         payload = data["deleteJob"]
         if payload.get("error"):
+            # A retry after a lost acknowledgement can legitimately receive a
+            # not-found mutation error.  The error alone is never release
+            # evidence; accept it only if a separate exact lookup proves the
+            # same ID absent or gone.
```
masalim2:
```suggestion
```

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -193,71 +243,189 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
...
+                return
             raise RuntimeError(f"GraphQL deleteJob failed:\n{payload['error']}")
+        returned_id = ((payload.get("node") or {}).get("jobId") or "").strip()
+        if returned_id and returned_id != job_id:
+            raise RuntimeError(
+                "GraphQL deleteJob returned a different job ID: "
+                f"expected {job_id!r}, got {returned_id!r}"
+            )
+
+        # A successful mutation is only an acknowledgement. Retain the DB
```
masalim2: What's the motivation to block until the job is fully cleaned up and terminal in the scheduler?  I get that it seems a bit cleaner logically, but it could bog down the controller if it's ever in a position where it needs to qdel a bunch of things all at once and the scheduler is taking a long time to respond.

Once the job gets a terminate signal, it's effectively on its way out, even if it takes some time to get cleaned up, so it's probably okay for our control plane to treat the job as killed even if there's a bit of async lag til the time of full job removal.  

Was there an area where this was causing issues with multinode on Tara?

masalim2 (commented):
This looks great.  Thanks for catching the missing pagination aspect: this is definitely more correct with the full listing, omitting history jobs, and failing closed if the pagination doesn't finish cleanly.  

My main question is around what motivated making the `terminate_job` blocking?  If we can get away with keeping that nonblocking I think things would still work correctly but I could be missing something that you encountered in testing with Tara multinode models.  Maybe there's another solution here too.

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
nscottnichols: With non-blocking, deleting marked the job deleted/gone as soon as `terminate_job()` returned. With multinode teardown, the scheduler needs more time for cleanup. I was hitting race condition issues when testing. Let me try to come up with something better than the block.

masalim2: It makes sense that the job needs some time for teardown, but that should be okay with the controller.  What types of bugs were the tests hitting?

My thinking is that from the controller's point of view, once a `qdel` is acknowledged, it can treat the batch job as gone and carry on.  In reality, the job might keep running for several minutes and tearing down, but the system shouldn't have to know about that or take any further actions; PBS will eventually tear everything down and free the nodes.  If the exiting job is preventing a _new job_ from starting in a timely fashion, that should just look like a queueing delay.  

I'm just wondering if there's some other issue that we're hitting that we couldn't address while keeping this simpler model of "qdel acknowledged; job is not my problem anymore".

nscottnichols: I thought more about this. Fixed and removed in [8526f39](https://github.com/argonne-lcf/FIRST/pull/282/commits/8526f39db6d3ffd948a4ebcb8ba4c03cb8b885cf).

After more discussion, actually fixed in [ff3a6d5](https://github.com/argonne-lcf/FIRST/pull/282/commits/ff3a6d5a5645db8b9dc7102665ebb26bc4119f79).

packages/gateway/first_gateway/platforms/schedulers/graphql_pbs.py
```diff
@@ -193,71 +240,160 @@ async def get_job_statuses(self) -> list[JobStatusInfo]:
...
+        if not edges:
+            return None
+        if len(edges) != 1 or not isinstance(edges[0], dict):
+            raise RuntimeError("GraphQL exact-job query returned non-exact edges")
+        edge = edges[0]
+        if edge.get("error"):
+            raise RuntimeError(f"GraphQL exact-job edge failed:\n{edge['error']}")
+        node = edge.get("node")
+        if not isinstance(node, dict):
+            raise RuntimeError("GraphQL exact-job edge returned no node")
```
masalim2:
```suggestion
        edge = data["jobs"]["edges"][0]
        if edge.get("error"):
            raise RuntimeError(f"GraphQL exact-job edge failed:\n{edge['error']}")
        node = edge["node"]
```

missed this one, but same feedback as for the list query simplification!

masalim2 (approved):

PR #283, opened 2026-08-17
Add multi-node PALS pilot inventory

nscottnichols:
## Summary

- Propagate requested node/GPU topology into the pilot runtime.
- Discover GPUs locally for single-node pilots and through one labeled PALS rank per host for multi-node pilots.
- Preserve scheduler rank order and validate hosts, ranks, GPU indices, counts, timeouts, and launcher results.
- Reject multi-node submissions without a configured PALS launcher.

## Why this is needed

SSH-based node discovery is unreliable on production HPC systems. FIRST must verify that observed topology matches the scheduler request before creating GPU claims or launching replicas.

## Tests

- Allocation, submitter, and replica-manager suites: 21 passed

### DIFF
--- a/packages/common/first_common/schema/pilot.py
+++ b/packages/common/first_common/schema/pilot.py
@@ -4,6 +4,7 @@ These schemas describe the communication between first-gateway and first-pilot.
 Do not confuse with admin-created pilot resources inside `resources` subpackage
 """
 
+import json
 import os
 from datetime import datetime
 from pathlib import Path
@@ -11,7 +12,7 @@ from tempfile import TemporaryDirectory
 from typing import Self
 
 import yaml
-from pydantic import BaseModel, PrivateAttr, computed_field
+from pydantic import BaseModel, Field, PrivateAttr, computed_field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from .types import GpuClaim, PilotLaunchSpec, ReplicaState
@@ -123,6 +124,9 @@ class PilotRuntimeConfig(BaseSettings):
     ip_allowlist: list[str]
     workdir: Path
     node_file_env: str
+    pals_path: Path | None = None
+    num_nodes: int = Field(ge=1)
+    gpus_per_node: int = Field(ge=1)
     job_name: str
 
     # Name of the IPv4 network interface (e.g. "hsn0") whose address the pilot
@@ -162,4 +166,35 @@ class PilotRuntimeConfig(BaseSettings):
         """
         yaml_path = os.environ["PILOT_CONFIG_FILE"]
         config_raw = yaml.safe_load(Path(yaml_path).read_text())
+        if not isinstance(config_raw, dict):
+            raise ValueError("pilot runtime config must be a YAML mapping")
+
+        # GraphQL-backed schedulers cannot stage a per-job YAML file. Overlay
+        # the submitted job's identity, topology, runtime paths, and allowlist;
+        # credentials and other static settings remain in the reviewed YAML.
+        for field_name in (
+            "job_name",
+            "external_port",
+            "nginx_path",
+            "workdir",
+            "node_file_env",
+            "pals_path",
+            "num_nodes",
+            "gpus_per_node",
+        ):
+            env_name = f"PILOT_{field_name.upper()}"
+            if env_name in os.environ:
+                value = os.environ[env_name]
+                config_raw[field_name] = None if value == "" else value
+
+        if "PILOT_IP_ALLOWLIST_JSON" in os.environ:
+            try:
+                allowlist = json.loads(os.environ["PILOT_IP_ALLOWLIST_JSON"])
+            except json.JSONDecodeError as exc:
+                raise ValueError("PILOT_IP_ALLOWLIST_JSON is invalid JSON") from exc
+            if not isinstance(allowlist, list) or not all(
+                isinstance(value, str) for value in allowlist
+            ):
+                raise ValueError("PILOT_IP_ALLOWLIST_JSON must be a string list")
+            config_raw["ip_allowlist"] = allowlist
         return cls.model_validate(config_raw)
--- a/packages/common/first_common/schema/pilot.py
+++ b/packages/common/first_common/schema/pilot.py
@@ -4,7 +4,6 @@ These schemas describe the communication between first-gateway and first-pilot.
 Do not confuse with admin-created pilot resources inside `resources` subpackage
 """
 
-import json
 import os
 from datetime import datetime
 from pathlib import Path
@@ -15,7 +14,13 @@ import yaml
 from pydantic import BaseModel, Field, PrivateAttr, computed_field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-from .types import GpuClaim, PilotLaunchSpec, ReplicaState
+from .types import (
+    GpuClaim,
+    GpuDiscovery,
+    PilotLaunchSpec,
+    ReplicaState,
+    SSHDiscovery,
+)
 
 
 class ReplicaStartRequest(BaseModel):
@@ -124,7 +129,7 @@ class PilotRuntimeConfig(BaseSettings):
     ip_allowlist: list[str]
     workdir: Path
     node_file_env: str
-    pals_path: Path | None = None
+    gpu_discovery: GpuDiscovery = Field(default_factory=SSHDiscovery)
     num_nodes: int = Field(ge=1)
     gpus_per_node: int = Field(ge=1)
     job_name: str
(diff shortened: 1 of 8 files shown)


tests/test_graphql_pbs_allocation.py
```diff
@@ -0,0 +1,132 @@
...
+            "scheduler_config": {},
+            "job_walltime_min": 90,
+            "queue": "workq",
+            "account": "inference_service",
+            "max_num_nodes": 2,
+            "gpus_per_node": 4,
+            "workdir": "/service/first-v2/workdir",
+            "external_port": 18443,
+            "nginx_path": "/service/nginx/sbin/nginx",
+            "ip_allowlist": ["10.124.176.33/32"],
```
lowpolyneko: It's probably worth subbing out the production cluster details with stubs. Since we're merely emulating successful responses in the testsuite just ensuring `data_in == data_out` is sufficent.

masalim2:
@nscottnichols can you provide some more rationale for this one?  I can definitely see that SSH is not guaranteed to work portably across HPC systems and we probably need a more general mechanism for querying GPUs across the allocation.  

However, it does seem to work on Sophia and Tara, and I worry that a full switchover to `pals` is also not ideal, because it's an HPE/Cray specific launch mechanism and I'd like to keep this portable to non-HPE systems.  Maybe we could have a configurable `gpu_query_fanout_method: "parallel_app_launcher" | "ssh"` that accommodates SSH or PALS or some other `mpirun` variant.  

I also saw that you had added the `num_nodes` and `gpus_per_node` directly to the startup configuration which is a nice call: continuing on that train of thought, I wonder if it's best to just use plain old indices like `gpus_ids = list(range(gpus_per_node))` without a mandatory discovery step.  

The automated `nvidia-smi` call is nice because it's cool to see the human-readable GPU names and memory usage in real time on the admin UI, but I wonder if being so strict on resource discovery is the right call.

packages/common/first_common/schema/pilot.py
```diff
@@ -123,6 +124,9 @@ class PilotRuntimeConfig(BaseSettings):
     ip_allowlist: list[str]
     workdir: Path
     node_file_env: str
+    pals_path: Path | None = None
```
masalim2: Let's consider an inventory method option that can select between different fanout methods like PALS and SSH, that way we're not moving from hardcoded `ssh` to hardcoded `pals`.

packages/common/first_common/schema/pilot.py
```diff
@@ -162,4 +166,35 @@ def load(cls) -> Self:
...
+        if "PILOT_IP_ALLOWLIST_JSON" in os.environ:
+            try:
+                allowlist = json.loads(os.environ["PILOT_IP_ALLOWLIST_JSON"])
+            except json.JSONDecodeError as exc:
+                raise ValueError("PILOT_IP_ALLOWLIST_JSON is invalid JSON") from exc
+            if not isinstance(allowlist, list) or not all(
+                isinstance(value, str) for value in allowlist
+            ):
+                raise ValueError("PILOT_IP_ALLOWLIST_JSON must be a string list")
+            config_raw["ip_allowlist"] = allowlist
```
masalim2:
```suggestion
```

This environment variable fallback discovery should already be working well and built-into the parent class `BaseSettings`.  Let's slim down the settings file on tara to just the PKI fields, and then the environment variables that you're passing through should get picked up nicely as `PILOT_*` variables automatically.

packages/gateway/first_gateway/services/pilot_submitter.py
```diff
@@ -62,6 +64,11 @@ async def submit(self, pilot_job: PilotJob | db.PilotJob) -> JobSubmitResult:
         scheduler_name = f"{pc.job_name_prefix}{name}"
         log_path = pc.workdir / "submit_scripts" / f"{name}.log"
 
+        if pilot_job.num_nodes > 1 and pc.pals_path is None:
+            raise ValueError(
+                "multi-node pilot submission requires PilotConfig.pals_path"
+            )
+
```
masalim2: This could be handled nicely if e.g. `PilotConfig.gpu_discovery` was a union type of `SSHDiscovery | PalsDiscovery`.  

Then pydantic will automatically validate that a PALS path is needed if the pilot config is using that path for GPU discovery, and it's a clean spot to extend if needed

packages/gateway/first_gateway/services/pilot_submitter.py
```diff
@@ -73,10 +80,26 @@ async def submit(self, pilot_job: PilotJob | db.PilotJob) -> JobSubmitResult:
                 raise ValueError(
                     "GraphQLPBSAdapter requires PilotConfig.pilot_config_path"
                 )
+            runtime_env = {
```
masalim2, line 78: Good call!  I like this approach; we can slim the config file on Tara down to everything that's not specified here (basically everything but the PKI secrets)

packages/pilot/first_pilot/replica_manager.py
```diff
@@ -36,15 +36,9 @@
...
-    try:
-        return pool.submit(socket.getfqdn, name).result(timeout=timeout)
-    except FutTimeout:
-        logger.warning("getfqdn(%r) timed out after %ss", name, timeout)
-        return name or socket.gethostname()
-    finally:
-        pool.shutdown(wait=False, cancel_futures=True)
+    """Return a DNS-free address label; retained for the control API contract."""
+    del timeout
+    return name or socket.gethostname()
```
masalim2: I think we just settled on using IP addresses with SSL `verify_hostname=False` across the board because ALCF DNS is on shaky ground.  So don't worry about the control API contract; let's just delete this whole function -- it's just cruft with a confusing/misleading name at this point.  

All the callers can be swapped out to an inline `socket.gethostname()` safely at this point.  We still get a nice human-readable hostname label but it doesn't need to be a fully-qualified DNS name.

packages/pilot/first_pilot/replica_manager.py
```diff
@@ -55,20 +49,35 @@ class _ReservedSentinel(Enum):
...
+
+
+def _require_executable(path: Path, *, label: str) -> str:
+    """Return a configured executable path or fail with a useful error."""
+    if not path.is_file() or not os.access(path, os.X_OK):
+        raise RuntimeError(f"required {label} is unavailable: {path}")
+    return str(path)
+
+
+def _parse_gpu_row(line: str) -> GpuInfo:
```
masalim2, line 62: `_parse_gpu_row` and `_validated_host_gpus` have lots of strict checks enabled.  

If we depend on this for correctly building the inventory of GPU ID's that are available at runtime, then I think that's the right call.  

But if we can leverage the new  `num_nodes` and `gpus_per_node` to build the inventory, then this logic could be demoted to a "nice-to-have; cool-to-see  in the admin UI", but it isn't in the critical path to correctly managing the GPUs in the allocation.  In that case I think the less strict posture of warning logs rather than fatal errors would be better.

tests/test_graphql_pbs_allocation.py
nscottnichols: Agreed! I'll rework this test.

nscottnichols:
@masalim2 I fully agree. For the issue I was hitting, it may have just been transient because they were working on the cluster (I was having problems with the SSH method). Another concern I had was if SSH access becomes more restricted, whether that would impact us. It sounds like you are in support of abstracting these concepts though which is what I was thinking about doing originally. Let me refactor this to address your concerns.

tests/test_graphql_pbs_allocation.py
nscottnichols: Fixed in [19d8bd1](https://github.com/argonne-lcf/FIRST/pull/283/commits/19d8bd1cee8f17207a453121ce29664476798789). I also removed the leaks in other tests.

packages/common/first_common/schema/pilot.py
nscottnichols: Fixed in [19d8bd1](https://github.com/argonne-lcf/FIRST/commit/19d8bd1cee8f17207a453121ce29664476798789).

Changed the runtime configuration field:

https://github.com/argonne-lcf/FIRST/blob/19d8bd1cee8f17207a453121ce29664476798789/packages/common/first_common/schema/pilot.py#L129-L134

Added the new `GpuDiscovery` type:

https://github.com/argonne-lcf/FIRST/blob/19d8bd1cee8f17207a453121ce29664476798789/packages/common/first_common/schema/types.py#L151-L171

packages/common/first_common/schema/pilot.py
nscottnichols: Fixed in [19d8bd1](https://github.com/argonne-lcf/FIRST/commit/19d8bd1cee8f17207a453121ce29664476798789).

Removed:

https://github.com/argonne-lcf/FIRST/blob/19d8bd1cee8f17207a453121ce29664476798789/packages/common/first_common/schema/pilot.py#L174-L177

packages/pilot/first_pilot/replica_manager.py
nscottnichols: Fixed in [19d8bd1](https://github.com/argonne-lcf/FIRST/commit/19d8bd1cee8f17207a453121ce29664476798789).

Removed the `safe_getfqdn()` function and replaced with the inline `sockets.gethostname()`.

packages/pilot/first_pilot/replica_manager.py
nscottnichols: Having the path here allows us to continue with a situation where maybe one GPU is broken on a node and allows us to recover/continue with a proper map to the resources. Failed parsing here shouldn't be a critical fail beyond we need to fix the code to handle those malformed cases. I did do a bit of refactoring here though (in anticipation of possibly other GPU architectures or AI accelerators).

packages/common/first_common/schema/types.py
```diff
@@ -172,6 +193,7 @@ class PilotConfig(BaseModel):
     nginx_path: Path
     ip_allowlist: list[str]
     node_file_env: str
+    gpu_discovery: GpuDiscovery = Field(default_factory=SSHDiscovery)
```
masalim2, line 196: Love it!

masalim2 (approved):

PR #285, opened 2026-08-17
Make replica teardown authoritative and retryable

nscottnichols:
## Summary

- Add bounded pre-stop and post-stop hooks.
- Report teardown success only after model and hook process groups are proven absent.
- Keep failed teardown attempts retryable while preventing signals to reused process-group IDs.
- Retain GPU claims until teardown succeeds.
- Bound stop-all and replica-stop execution.

## Why this is needed

A model leader can exit while child processes continue holding GPUs. Releasing claims at that point can place new work onto resources that are still occupied.

## Tests

- Replica lifecycle, manager lifecycle, and control suites passed
- 52 focused tests pass cumulatively through this stack

### DIFF
--- a/packages/pilot/first_pilot/replica.py
+++ b/packages/pilot/first_pilot/replica.py
@@ -6,7 +6,9 @@ import subprocess
 import threading
 import time
 from datetime import datetime, timezone
+from enum import Enum, auto
 from pathlib import Path
+from typing import NoReturn
 from urllib.parse import urlparse
 
 from httpx import Client, HTTPTransport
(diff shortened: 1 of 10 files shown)


packages/pilot/first_pilot/replica.py
```diff
@@ -130,7 +154,21 @@ def __init__(
...
+        self._pre_stop_succeeded: bool | None = None
+        self._post_stop_proc: subprocess.Popen[bytes] | None = None
+        self._post_stop_succeeded: bool | None = None
+        self._post_stop_attempts = 0
+        self._model_teardown_attempts = 0
+        self._hook_cleanup_attempts = 0
+        self._post_hook_cleanup_attempts = 0
+        self._model_survivors: dict[int, int] | None = None
+        self._hook_survivors: dict[int, int] | None = None
+        self._post_hook_survivors: dict[int, int] | None = None
```
lowpolyneko: I think it'd be worth storing the hook states in a separate struct to improve readability, each phase more or less stores the same information.

packages/pilot/first_pilot/replica_manager.py
```diff
@@ -385,10 +388,21 @@ def _allocate_uds_locked(self) -> str:
...
+        name: str,
+        resources: list[GpuClaim],
+        *,
+        expected: Replica | ReservedSentinel,
+    ) -> bool:
+        """Release only if ``name`` still refers to the exact owned object."""
+        # Caller must hold self._lock. The identity guard prevents a late
+        # concurrent stop from releasing a replacement Replica's claims.
+        if self._replicas.get(name) is not expected:
+            return False
```
masalim2: I'm not sure if the new `expected` argument and identity guard is totally necessary, because the replica `name` should be enforced unique on the backend; a replacement replica should get started with a new unique name.

packages/pilot/first_pilot/replica_manager.py
```diff
@@ -449,31 +467,104 @@ def stop_replica(self, replica_name: str) -> None:
...
+                expected=replica,
+            )
+        if not released:
+            logger.info(
+                "Replica %s stopped, but its manager entry now has a different "
+                "identity; leaving the current entry untouched",
+                replica_name,
+            )
 
     def stop_all(self) -> None:
```
masalim2: Looks good!  I would add that `stop_all` should only be called when the whole PBS job is on its way out and no more Replicas should get placed on it.  So even if shutdown is fully clean, it should be okay as long as the scheduler cleans up properly on job exit.

packages/pilot/first_pilot/replica.py
```diff
@@ -182,18 +220,35 @@ def _unhealthy_for_too_long(self) -> bool:
...
         while not self._monitor_exit.wait(timeout=self._HEALTH_INTERVAL):
             try:
                 self._run_monitor_check()
+            except ReplicaTeardownError:
+                # The teardown path already recorded exact surviving groups in
+                # state_message and the lifecycle log.  Keep the Replica
+                # addressable so a later controller retry can finish cleanup.
+                logger.exception("teardown incomplete for replica %s", self.name)
+                if self.state != ReplicaState.terminating:
+                    self.state = ReplicaState.error
```
masalim2: Let's add a note to `self.state_message` here to surface why the state transitioned to error.  
```suggestion
                    self.state = ReplicaState.error
                    self.state_message = f"Encountered teardown error: {exc}"
```

packages/pilot/first_pilot/replica.py
```diff
@@ -182,18 +220,35 @@ def _unhealthy_for_too_long(self) -> bool:
...
+
+    def _unhealthy_timeout_sec(self) -> int:
+        timeout = self.launch_spec.max_unhealthy_sec
+        return self.launch_spec.max_startup_sec if timeout is None else timeout
 
     def _monitor_loop(self) -> None:
         while not self._monitor_exit.wait(timeout=self._HEALTH_INTERVAL):
             try:
                 self._run_monitor_check()
+            except ReplicaTeardownError:
```
masalim2:
```suggestion
            except ReplicaTeardownError as exc:
```

masalim2 (commented):
(Still working through a commit; leaving a few smaller comments now)

packages/pilot/first_pilot/replica_manager.py
```diff
@@ -503,31 +522,104 @@ def stop_replica(self, replica_name: str) -> None:
...
         replica.stop()
 
         with self._lock:
-            self._release_locked(replica_name, replica.resources)
+            released = self._release_locked(
+                replica_name,
+                replica.resources,
+                expected=replica,
+            )
+        if not released:
```
masalim2, line 533: same logic as the comment above; replica `name` is enforced unique inside postgres before it makes it out here; can omit this guard

packages/pilot/first_pilot/replica_manager.py
```diff
@@ -503,31 +522,104 @@ def stop_replica(self, replica_name: str) -> None:
...
+                daemon=True,
+            )
+            try:
+                worker.start()
+            except BaseException as exc:
+                errors[id(replica)] = exc
+                done_events[id(replica)].set()
+                logger.exception(
+                    "could not start stop_all worker for replica %s", replica.name
+                )
```
masalim2:
```suggestion
            worker.start()
```

Can be less defensive; if a thread fails to start we probably have bigger problems; can  let this crash and diagnose.

packages/pilot/first_pilot/replica.py
```diff
@@ -24,6 +26,124 @@
 logger = logging.getLogger(__name__)
 
 
+class ReplicaTeardownError(RuntimeError):
```
masalim2: This is actually worth moving into `first_common/errors.py` as a sublcass of `FirstError`.  This ensures that the exception handler in `control_api.py` picks up the error and returns a nicely-formatted JSON response to the controller instead of just `Internal server error`.  

Then we should see it as an informative Replica reconcile error in the admin UI.

nscottnichols:
Key changes:

  - Added deterministic hook-ownership regression. It failed before the fix because _pre_stop_group remained None after launch succeeded but wait() raised.
  - Store hook groups immediately after launch.
  - Prevent retries from launching another post-stop verifier while the prior group remains.
  - Moved ReplicaTeardownError into first_common.errors as HTTP 409 / replica_teardown_incomplete.
  - Propagate structured pilot teardown details into gateway reconcile_last_error, with plain-text fallback.
  - Simplified replica release logic while preserving duplicate-release safety.
  - Replaced the same-name replacement test with the realistic A/B GPU-claim scenario.

PR #286, opened 2026-08-17
Verify pilot readiness and account pre-manager failures

nscottnichols:
## Summary

- Confirm GraphQL candidate endpoints through the pilot control API before publishing readiness.
- Use exact scheduler lookup when a job is absent from the bulk listing.
- Atomically count pilot allocations that terminate before manager readiness.
- Exclude intentional deletion and ensure each failure is counted once.

## Why this is needed

The scheduler can expose a head-node address before the pilot validates its resources or binds its control API. Publishing that address early causes premature launches, while uncounted startup failures can create an unlimited replacement loop.

## Tests

- All 9 observer tests collect successfully
- Database execution is pending CI because local PostgreSQL/Redis could not be started

### DIFF
--- a/packages/gateway/first_gateway/controllers/workers/pilot_job_observer.py
+++ b/packages/gateway/first_gateway/controllers/workers/pilot_job_observer.py
@@ -125,11 +125,8 @@ class PilotJobObserver(Worker):
         for job in db_jobs:
             assert job.scheduler_job_id is not None
             status = statuses.get(job.scheduler_job_id)
-            if status is None and isinstance(submitter.adapter, GraphQLPBSAdapter):
+            if status is None:
                 # Bulk scheduler listings can be truncated or active-only.
-                # Absence is lifecycle evidence only when the adapter's exact-ID
-                # lookup explicitly proves it. Other adapters retain their
-                # existing missing-means-gone behavior.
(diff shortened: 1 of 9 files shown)


lowpolyneko (approved):

packages/gateway/first_gateway/controllers/workers/pilot_job_observer.py
```diff
@@ -125,49 +159,101 @@ async def _poll_cluster(
...
             logger.info(
-                f"PilotJob {db_job.name}: {db_job.scheduler_state} -> {status.state.value}"
+                f"PilotJob {db_job.name}: {db_job.scheduler_state} -> {target_state}"
+            )
+
+    @staticmethod
+    async def _record_pre_manager_launch_failure(
+        sess: AsyncSession, job: PilotJob
+    ) -> None:
+        """Charge each actively assigned deployment once for a dead pilot.
```
masalim2, line 206: It doesn't seem fair to ascribe pilot startup failures to the associated deployments that never had a chance to launch onto the pilot. 

I do agree it makes sense to backoff on `qsub`/`qstat` if something is going wrong with the scheduler or `first-pilot`.  Perhaps we can do that separately (as a cluster-level counter, or counting it as a cluster reconcile failure).

PR #287, opened 2026-08-17
Guard gateway allocation lifecycle races

nscottnichols:
## Summary

- Recheck placement eligibility while holding the pilot-job lock.
- Prevent idle deletion when a live replica has been assigned concurrently.
- Keep exiting jobs in capacity accounting until exact scheduler absence.
- Retain GPU claims after failed manager cleanup and escalate repeated failures to parent allocation termination.
- Finalize replica deletion only after the parent allocation is gone.

## Why this is needed

Concurrent controllers can otherwise assign a replica while another controller deletes its pilot, or release GPU claims before cleanup is proven complete. These guards keep placement, draining, and scheduler termination consistent.

## Tests

- All 60 controller, drainer, and placement tests collect successfully
- Database execution is pending CI because local PostgreSQL/Redis could not be started

### DIFF
--- a/packages/gateway/first_gateway/controllers/workers/pilot_job_controller.py
+++ b/packages/gateway/first_gateway/controllers/workers/pilot_job_controller.py
@@ -8,7 +8,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from first_common.schema.base_scheduler import JobSubmitResult, SchedulerJobState
 from first_common.schema.types import HealthCheckResult, PilotConfig
 
-from ...database.models import Cluster, PilotJob
+from ...database.models import Cluster, PilotJob, PilotReplica
 from ...database.redis.pubsub import Channel
 from ...platforms.schedulers import build_scheduler
 from ...services.pilot_submitter import PilotSubmitter
@@ -46,7 +46,10 @@ class PilotJobController(Controller):
                         SchedulerJobState.running.value,
                     ]
                 ),
-                PilotJob.idle_since.is_not(None),
+                sa.and_(
+                    PilotJob.idle_since.is_not(None),
+                    self._no_live_assigned_replicas(),
+                ),
                 PilotJob.manager_health == HealthCheckResult.unhealthy.value,
             ),
         )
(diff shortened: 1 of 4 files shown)


packages/gateway/first_gateway/controllers/workers/replica_drainer.py
```diff
@@ -102,17 +117,117 @@ async def reconcile(self, uid: int) -> None:
...
+        # absence is the only safe release proof.  In particular, the forced
+        # cleanup path must not unassign claims while qdel is still running.
+        if job and job.scheduled_deletion_at is not None:
+            if (
+                job.scheduler_state != SchedulerJobState.gone.value
+                or job.deleted_at is None
+            ):
+                return
+            await self._finalize_deletion(replica)
+            return
```
lowpolyneko:
```suggestion
            if (
                job.scheduler_state == SchedulerJobState.gone.value
                and job.deleted_at is not None
            ):
                await self._finalize_deletion(replica)
            return
```

lowpolyneko (approved):
Looks good otherwise.

masalim2:
Can you help me understand if you encountered some issues that necessitate these changes?

> Concurrent controllers can otherwise assign a replica while another controller deletes its pilot, or release GPU claims before cleanup is proven complete. These guards keep placement, draining, and scheduler termination consistent.

I thought about concurrent controllers and went back-and-forth quite a bit at the onset of this.  I ultimately decided to hold off on parallel controllers until we hit a real bottleneck requiring them, and therefore to avoid `SELECT FOR UPDATE` as much as possible by making the controller a singleton (the `ManagerLease` enforced at startup was intended to help us avoid accidentally ever running two controllers at the same time and sidestep concurrency issues).  

>     Recheck placement eligibility while holding the pilot-job lock.

Holding DB row locks while doing RPCs can cause some contention issues; it's better to set up something like a transactional outbox pattern when the need really arises to have concurrent workers.  It shouldn't be an issue in the current design.

> Prevent idle deletion when a live replica has been assigned concurrently.

Good call, that's definitely a race that can happen even with one controller.

> Keep exiting jobs in capacity accounting until exact scheduler absence.

Sounds good! 

> Retain GPU claims after failed manager cleanup and escalate repeated failures to parent allocation termination.

This should already be the case: it won't release the claims if it gets an error from the `/stop-replica` API.

PR #288, opened 2026-08-18
ci: add workflow for alcf-ai

lowpolyneko:
Automatically publishes the `alcf-ai` package on version tag creation or manual trigger.

Requires this repository to be added as a [Trusted Publisher](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/).

### DIFF
--- a/.github/workflows/publish-alcf-ai.yml
+++ b/.github/workflows/publish-alcf-ai.yml
@@ -1,0 +1,36 @@
+name: Publish alcf-ai to PyPI
+
+on:
+  push:
+    tags:
+      - 'alcf_ai/*'
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


masalim2:
image

Thanks for this! If we use `git push --tags` as the trigger, should we also remove `version` from pyproject.toml and configure a build plugin like https://github.com/ofek/hatch-vcs to dynamically get version from git?  

It's not a big deal to just manually keep these in sync but kind of nice to have a single source of truth from the git tag, up to you!

masalim2 (approved):

lowpolyneko:
Good idea! I did just that :)

PR #303, opened 2026-08-20
deploy/gateway: update service file to write logs to /home

lowpolyneko:
Reflection of prod changes.

@bcote-anl we should update the backup cronjob to pull from the new log storage directory aswell!


bcote-anl (approved):
Thanks!

PR #305, opened 2026-08-21
health_monitor: pull the executor from cache before every health check

lowpolyneko:
In the event the Globus executor encounters a `RequestTimeout` whilst performing a health check, the executor is shutdown, causing all subsequent health checks to fail. Instead, we should re-pull the executor from the cache prior to each health check.

I believe this is the only case where an Executor is reused after submitting a task that can potentially time out in the entire codebase.


lowpolyneko:
I made the change on prod and I'll let it run abit prior to merging to make sure the transient timeout issues are resolved.

bcote-anl (approved):

PR #306, opened 2026-08-21
Document and test public model capability metadata

nscottnichols:
## Summary

Adds documentation and regression tests for exposing client-neutral model metadata through FIRST’s existing authenticated /models endpoint.

Key behavior changes

- Allows display_name, description, and capabilities when explicitly configured in MODEL_DETAILS_KEYS.
- Preserves model-level authorization.
- Prevents backend URLs, credentials, and certificate paths from being returned.
- Models without the new metadata remain unchanged.

## Testing
All 252 FIRST tests, Ruff, Mypy, and the strict documentation build passed.


lowpolyneko (approved):
LGTM, thanks for documenting this and adding a test harness to prevent leaks!

PR #307, opened 2026-08-21
Firstv2 httpx

bcote-anl:
Adding a backend client manager to create an httpx AsyncClient for each healthy backend. The PR includes modification on the RouterConfig manager to allow other classes to register to changes in Redis, which represents functions to be executed following the `_swap` function. When models are updates in Redis, the BackendClientManager should automatically refresh its httpx clients.

### DIFF
--- a/packages/gateway/first_gateway/apiserver/api.py
+++ b/packages/gateway/first_gateway/apiserver/api.py
@@ -26,7 +26,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     config_logging(settings.log_level)
     async with settings.build_clients() as client_state:
         app.state.client_state = client_state
-        backend_client_manager = BackendClientManager()
+        backend_client_manager = BackendClientManager(client_state.settings)
         router_config_manager = RouterConfigManager(client_state.redis)
         router_config_manager.add_swap_callback(backend_client_manager.on_config_swap)
         await router_config_manager.start()
--- a/packages/gateway/first_gateway/apiserver/api.py
+++ b/packages/gateway/first_gateway/apiserver/api.py
@@ -7,6 +7,7 @@ from fastapi import FastAPI
 from first_gateway.settings import Settings
 
 from ..log_config import config_logging
+from .backend_client_manager import BackendClientManager
 from .error_handlers import register_error_handlers
 from .log_middleware import log_request
 from .router_config_manager import RouterConfigManager
@@ -25,13 +26,17 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     config_logging(settings.log_level)
     async with settings.build_clients() as client_state:
         app.state.client_state = client_state
+        backend_client_manager = BackendClientManager()
         router_config_manager = RouterConfigManager(client_state.redis)
+        router_config_manager.add_swap_callback(backend_client_manager.on_config_swap)
         await router_config_manager.start()
         app.state.router_config_manager = router_config_manager
+        app.state.backend_client_manager = backend_client_manager
         try:
             yield
         finally:
             await router_config_manager.stop()
+            await backend_client_manager.close_all()
 
 
 app = FastAPI(title="ALCF Inference Service", lifespan=lifespan)
--- a/packages/gateway/first_gateway/apiserver/backend_client_manager.py
+++ b/packages/gateway/first_gateway/apiserver/backend_client_manager.py
@@ -11,9 +11,17 @@ from ..database.redis.router_config import BackendConfig, DeploymentConfig, Rout
 from ..services.certmanager import generate_client_cert
 from ..settings import Settings
 
+# To accomodate different OS platform (MacOS/Linux)
+if hasattr(socket, "TCP_KEEPIDLE"):
+    TCP_KEEPIDLE_OR_KEEPALIVE = socket.TCP_KEEPIDLE
+elif hasattr(socket, "TCP_KEEPALIVE"):
+    TCP_KEEPIDLE_OR_KEEPALIVE = socket.TCP_KEEPALIVE
+else:
+    TCP_KEEPIDLE_OR_KEEPALIVE = None
+
 SOCKET_OPTS = [
     (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),
-    (socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, 60),
+    (socket.IPPROTO_TCP, TCP_KEEPIDLE_OR_KEEPALIVE, 60),
     (socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 15),
     (socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 4),
 ]
--- a/packages/gateway/first_gateway/apiserver/backend_client_manager.py
+++ b/packages/gateway/first_gateway/apiserver/backend_client_manager.py
@@ -1,3 +1,4 @@
+import asyncio
 import ssl
 import tempfile
 from pathlib import Path
@@ -47,9 +48,10 @@ class BackendClientManager:
 
         # Backends that should be removed
         for backend_id in set(self._clients) - incoming.keys():
-            await self._close_backend(backend_id)
+            await self._close_backend(backend_id, sleep=60)
 
-    async def _close_backend(self, backend_id: str) -> None:
+    async def _close_backend(self, backend_id: str, sleep: int = 0) -> None:
+        await asyncio.sleep(sleep)
         client = self._clients.pop(backend_id)
         self._configs.pop(backend_id)
         if client:
--- a/packages/gateway/first_gateway/apiserver/backend_client_manager.py
+++ b/packages/gateway/first_gateway/apiserver/backend_client_manager.py
@@ -52,7 +52,9 @@ class BackendClientManager:
     async def _close_backend(self, backend_id: str) -> None:
         client = self._clients.pop(backend_id)
         self._configs.pop(backend_id)
-        await client.aclose()
+        if client:
+            await client.aclose()
+        # TODO: raise a warning here if client is None?
 
     async def close_all(self) -> None:
         for client in list(self._clients.values()):
--- a/packages/gateway/first_gateway/apiserver/backend_client_manager.py
+++ b/packages/gateway/first_gateway/apiserver/backend_client_manager.py
@@ -12,12 +12,10 @@ from ..services.certmanager import generate_client_cert
 from ..settings import Settings
 
 # To accomodate different OS platform (MacOS/Linux)
-if hasattr(socket, "TCP_KEEPIDLE"):
-    TCP_KEEPIDLE_OR_KEEPALIVE = socket.TCP_KEEPIDLE
-elif hasattr(socket, "TCP_KEEPALIVE"):
+if hasattr(socket, "TCP_KEEPALIVE"):
     TCP_KEEPIDLE_OR_KEEPALIVE = socket.TCP_KEEPALIVE
 else:
-    TCP_KEEPIDLE_OR_KEEPALIVE = None
+    TCP_KEEPIDLE_OR_KEEPALIVE = socket.TCP_KEEPIDLE
 
 SOCKET_OPTS = [
     (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),
--- a/packages/gateway/first_gateway/apiserver/backend_client_manager.py
+++ b/packages/gateway/first_gateway/apiserver/backend_client_manager.py
@@ -20,6 +20,7 @@ class BackendClientManager:
     def __init__(self, settings: Settings) -> None:
         self._settings = settings
         self._clients: dict[str, httpx.AsyncClient] = {}
+        self._configs: dict[str, tuple[BackendConfig, DeploymentConfig]] = {}
 
     def get(self, backend_id: str) -> httpx.AsyncClient | None:
         return self._clients.get(backend_id)
@@ -32,23 +33,32 @@ class BackendClientManager:
             for backend in deployment.backends
         }
 
-        current_ids = set(self._clients)
-
+        # Backends that should be added or updated
         for backend_id, (backend, deployment) in incoming.items():
-            # New backends
-            if backend_id not in current_ids:
+            if backend_id in self._clients and (
+                backend,
+                deployment,
+            ) != self._configs.get(backend_id):
+                # Remove existing clients that require an update
+                await self._close_backend(backend_id)
+            if backend_id not in self._clients:
                 self._clients[backend_id] = self._create_client(backend, deployment)
-            # TODO: Do something if existing backend had some properties changed?
+                self._configs[backend_id] = (backend, deployment)
 
-        # Deprecated backends
-        for backend_id in current_ids - incoming.keys():
-            client = self._clients.pop(backend_id)
-            await client.aclose()
+        # Backends that should be removed
+        for backend_id in set(self._clients) - incoming.keys():
+            await self._close_backend(backend_id)
+
+    async def _close_backend(self, backend_id: str) -> None:
+        client = self._clients.pop(backend_id)
+        self._configs.pop(backend_id)
+        await client.aclose()
 
     async def close_all(self) -> None:
         for client in list(self._clients.values()):
             await client.aclose()
         self._clients.clear()
+        self._configs.clear()
 
     def _create_client(
         self, backend: BackendConfig, deployment: DeploymentConfig
--- a/packages/gateway/first_gateway/apiserver/backend_client_manager.py
+++ b/packages/gateway/first_gateway/apiserver/backend_client_manager.py
@@ -1,4 +1,5 @@
 import asyncio
+import socket
 import tempfile
 from pathlib import Path
 from ssl import SSLContext, create_default_context
@@ -9,6 +10,18 @@ import httpx
 from ..database.redis.router_config import BackendConfig, DeploymentConfig, RouterConfig
 from ..services.certmanager import generate_client_cert
 from ..settings import Settings
+
+SOCKET_OPTS = [
+    (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1),
+    (socket.IPPROTO_TCP, socket.TCP_KEEPALIVE, 60),
+    (socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 15),
+    (socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 4),
+]
+
+STREAM_TIMEOUT = httpx.Timeout(connect=5.0, read=120.0, write=60.0, pool=5.0)
+UNARY_TIMEOUT = httpx.Timeout(connect=5.0, read=900.0, write=60.0, pool=5.0)
+
+KEEPALIVE_EXPIRY = 30.0
 
 
 class BackendClientManager:
@@ -58,7 +71,6 @@ class BackendClientManager:
                 backend,
                 deployment,
             ) != self._configs.get(backend_id):
-                # Remove existing clients that require an update
                 await self._close_client(backend_id, sleep=0)
             if backend_id not in self._clients:
                 self._clients[backend_id] = self._create_client(backend, deployment)
@@ -74,7 +86,6 @@ class BackendClientManager:
         self._configs.pop(backend_id)
         if client:
             await client.aclose()
-        # TODO: raise a warning here if client is None?
 
     async def close_all(self) -> None:
         for client in list(self._clients.values()):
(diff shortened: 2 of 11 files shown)


packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
```diff
@@ -23,13 +24,16 @@ async def chat_completions(
...
+    backend_client_manager: BackendClientManagerDep,
     payload: OpenAIChatCompletionsPayload,
 ) -> StreamingResponse | dict[str, Any]:
     return await submit_inference_with_retry(
         user,
         model,
         admission_controller,
+        backend_client_manager,
         payload,
+        endpoint=OpenAIEndpoints.chat_completions,
```
masalim2: Since we use a distinct pydantic model for each endpoint like `OpenAIChatCompletionsPayload`, consider setting the endpoint directly on the model as a ClassVar ([automatically ignored by pydantic](https://pydantic.dev/docs/validation/dev/concepts/models/#class-variables)):

```python
class OpenAIChatCompletionsPayload(BasePayload):
    endpoint: ClassVar[Literal[OpenAIEndpoints.chat_completions]] = OpenAIEndpoints.chat_completions
```

With a `Literal` type it might be less important to use the Enum:

```python
class OpenAIChatCompletionsPayload(BasePayload):
    endpoint: ClassVar[Literal["chat/completions"]] = "chat/completions"
```

The parent class could express this as:

```python
class BasePayload(BaseModelAllowExtra):
    endpoint: ClassVar[str]

    def __init_subclass__(cls, **kwargs: Any) -> None:
        super().__init_subclass__(**kwargs)
        if not getattr(cls, "endpoint", None):
            raise TypeError(
                f"{cls.__name__} must set a non-empty `endpoint` class variable"
            )
```

The benefit of this is mostly just avoiding having to carry around the endpoint alongside the payload object, since payload is already specific to the endpoint.  It won't interfere with schema validation or payload serialization while allowing you to write:

```python
    response = await client.post(
        f"/v1/{payload.endpoint}",
        json=payload.model_dump(exclude_unset=True, mode="json"),
    )
```

packages/gateway/first_gateway/apiserver/routes/openai/federated.py
```diff
@@ -21,13 +22,16 @@ async def chat_completions(
...
+    backend_client_manager: BackendClientManagerDep,
     payload: OpenAIChatCompletionsPayload,
 ) -> StreamingResponse | dict[str, Any]:
     return await submit_inference_with_retry(
         user,
         model,
         admission_controller,
+        backend_client_manager,
         payload,
+        endpoint=OpenAIEndpoints.chat_completions,
```
masalim2: Same suggestion as openai/deployments.py

packages/gateway/first_gateway/apiserver/api.py
```diff
@@ -25,13 +26,17 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     config_logging(settings.log_level)
     async with settings.build_clients() as client_state:
         app.state.client_state = client_state
+        backend_client_manager = BackendClientManager(client_state.settings)
         router_config_manager = RouterConfigManager(client_state.redis)
+        router_config_manager.add_swap_callback(backend_client_manager.on_config_swap)
```
masalim2, line 31: This swap callback is a perfectly viable approach. I would also weigh it against a design where the client is lazy-loaded on the first request.   That might slightly simplify the code if everything is happening in the request pathway as opposed to a background task, but I would need to think about it more (and read the rest of this PR)

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,88 @@
...
+            for backend in deployment.backends
+        }
+
+        current_ids = set(self._clients)
+
+        for backend_id, (backend, deployment) in incoming.items():
+            # New backends
+            if backend_id not in current_ids:
+                self._clients[backend_id] = self._create_client(backend, deployment)
+            # TODO: Do something if existing backend had some properties changed?
```
masalim2: Great question!  Fortunately stuff like `weight` in the load balancing configuration gets refreshed and your admission control logic always takes the latest one.  What really could impact the AsyncClient is:

- base url
- headers
- SSL certificates

For a Pilot Deployment, all 3 of these should be stable for a given `backend_id`.  If something is changed in the configuration that impacts these items, we would have to spin up new replicas that would have a new PK and therefore a different/new `backend_id`.

For a StaticDeployment, I could totally imagine that the API key or Base URL changes.  So yes: to be robust we could have an `_update_client()` that ensures the Base URL and Headers are in sync.  This side of the implementation shouldn't care about the type of deployment or what is possible to change, so I think generically sync'ing existing clients is probably a good idea.

packages/gateway/first_gateway/apiserver/api.py
```diff
@@ -25,13 +26,17 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
...
         await router_config_manager.start()
         app.state.router_config_manager = router_config_manager
+        app.state.backend_client_manager = backend_client_manager
         try:
             yield
         finally:
             await router_config_manager.stop()
+            await backend_client_manager.close_all()
```
masalim2, line 39: This is really clean, I like how it fits on the app lifecycle! ⭐

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Deprecated backends
+        for backend_id in current_ids - incoming.keys():
+            client = self._clients.pop(backend_id)
+            await client.aclose()
```
masalim2, line 126: This cleanup pattern is very nice.  The one other thought I had is that we should delay `aclose()` for ~60 seconds after the backend disappears from the RouterConfig to allow any in-flight request to finish before closing the connections.  

The `RouterConfigObserver` [immediately removes a backend from Redis](https://github.com/argonne-lcf/FIRST/blob/firstv2/packages/gateway/first_gateway/controllers/workers/router_config_observer.py#L99) when the replica is marked for removal, so it's immediately invisible to new requests. 

However, to avoid killing vLLM in the middle of generation, the `ReplicaDrainer` [waits at least 20 seconds](https://github.com/argonne-lcf/FIRST/blob/firstv2/packages/gateway/first_gateway/controllers/workers/replica_drainer.py#L135) before terminating the replica.   We could mirror it on this side by wrapping the logic in a cleanup method that just delays a bit, then schedule the removal as an asyncio task `asyncio.create_task`:

```python
await asyncio.sleep(60)
client = self._clients.pop(backend_id, None)
if client:
    await client.aclose()
```

The only other detail to manage with background asyncio tasks is that they won't reliably run if garbage-collected, so persist the cleanup task on the class.  (This is the same issue motivating the `_background_tasks` in [log_middleware.py](https://github.com/argonne-lcf/FIRST/blob/firstv2/packages/gateway/first_gateway/apiserver/log_middleware.py#L90)).  

The 1 minute delay is probably much simpler and more performant than trying to maintain a consistent counter of users per client and will cover >99% of issues.  You could even extend it to a 5 minute delay or something generous.  The task always eventually runs and should  solve the main problem of a memory leak from several **days** of accumulating dead clients.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,88 @@
...
+                cn=backend.id,
+                ca_cert_pem=settings.pilot_ca_crt,
+                ca_key_pem=settings.pilot_ca_key.get_secret_value(),
+            )
+            with tempfile.TemporaryDirectory(delete=True) as tmpdir:
+                crt_path = Path(tmpdir) / "client.crt"
+                key_path = Path(tmpdir) / "client.key"
+                crt_path.write_text(client_crt_pem)
+                key_path.write_text(client_key_pem)
+                ctx.load_cert_chain(crt_path, key_path)
```
masalim2: This whole chunk of preparing the SSL context should be the same for all pilot deployments, so what if we just do this work one time in `__init__` and store `self.ctx` for future re-use?  

By default `generate_client_cert()` generates a certificate that lasts for 2 years; this indefinitely stored ctx would only be an issue if we ran the same server process continually for that long without restarting it.  It's probably safe to assume that's not an issue but we could always be extra-defensive and use `ctx = self.get_ssl_context()` to refresh the certificate if it's going to expire soon.  Probably overengineering at this point.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,88 @@
...
+            return httpx.AsyncClient(
+                base_url=backend.model_url,
+                headers=headers,
+                verify=ctx,
+            )
+
+        return httpx.AsyncClient(
```
masalim2, line 157: This looks great and handles all of the most important considerations for client authentication.  

Since this PR establishes the client builder, before we merge let's also just revisit any other client configuration that could be useful at this level: https://github.com/argonne-lcf/FIRST/blob/firstv2-httpx/docs/architecture/nginx-httpx.md#client-configuration .  

Setting up proxy behavior can probably be skipped: we can continue to use the blanket `ALL_PROXY` in `deploy/compose.tunnel.yaml` for development.

packages/gateway/first_gateway/apiserver/router_config_manager.py
```diff
@@ -25,19 +32,33 @@ def __init__(self, redis: Redis) -> None:
...
+    async def _swap(self, cfg: RouterConfig) -> None:
         # Both drivers read Redis independently; guard against a slow in-flight
         # load overwriting a newer config that already landed.
         if cfg.version >= self._current.version:
             self._current = cfg
+            for cb in self._swap_callbacks:
+                try:
+                    await cb(cfg)
+                except Exception:
+                    # TODO: Find a better way to handle errors here (might need to raise exception)
```
masalim2: I think this is perfectly fine, since it's running in the background, we don't have an exception to bubble up to anyone besides the logs.  

The one thing we could possibly do is start to elaborate our `/health` API endpoint from just `"ok"` to perhaps reflect the status of this engine?  Then if it's failing, we could flip a flag to get Slack health alerts that this part of the gateway is failing (and go look at the logs).  Seems like a nice minor enhancement for a future PR

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -41,7 +45,7 @@ async def submit_inference_with_retry(
     estimated_tokens = len(payload.model_dump()) // 2
 
     for attempt in range(min(3, len(backend_candidates))):
-        backend = await get_backend(
+        backend_id = await get_backend_id(
```
masalim2: I know this isn't part of the current PR, but now that I am re-reading, I have a small naming request purely for maintainability.  

Can we change `get_backend_id`  (which reads like a pure "read-only" action) to something like `admit_request()` to help readers remember that this is mutating request admission state in Redis?

bcote-anl:
Not sure why the test is failing. `make test` worked on my laptop and this succeeds 
tests/test_health_alerter.py ...................                                                                                         [ 33%]
which is the one failing on the Github CI

masalim2 (commented):
This looks great!  I left a comment debating "swap callback" versus "lazy load client".  

I think your swap logic is working great, on second thought.  If we tried to  build/cache the clients during `.get(backend_id)` on-demand, we would still need a way to 1) catch any updates in Redis if the client already exists and 2) cleanup old clients.  

So your approach seems cleaner to me, overall!  

I do think we should handle the case of updating client config and delaying client cleanup as suggested on the code comments, to get this prepared for production.  There's a handful of important additional client settings (timeouts; connection keepalive) that we can configure as well.  

I believe you nailed the design of the `BackendClientManager` and its wiring into the application most importantly, thank you!  

I do agree that we can revisit/refactor the exact shape of `submit_inference()` to handle different types of requests, error handling, etc...  That can fit naturally in with a future PR!

packages/gateway/first_gateway/apiserver/api.py
masalim2: (After reading the rest of the PR and thinking more, I think I like your original swap callback design better, because it keeps everything together: `create new client / handle update to existing client / cleanup drained client`)

masalim2:
> Not sure why the test is failing. `make test` worked on my laptop and this succeeds tests/test_health_alerter.py ................... [ 33%] which is the one failing on the Github CI

That is my fault, @bcote-anl -- I introduced a failing test case on `firstv2` and I think GitHub runs the checks in the _merged_ environment if it can (if you rebased you should see the new failure) .  I will fix, it's on me!

masalim2:
> > Not sure why the test is failing. `make test` worked on my laptop and this succeeds tests/test_health_alerter.py ................... [ 33%] which is the one failing on the Github CI
> 
> That is my fault, @bcote-anl -- I introduced a failing test case on `firstv2` and I think GitHub runs the checks in the _merged_ environment if it can (if you rebased you should see the new failure) . I will fix, it's on me!

@bcote-anl fixed!

packages/gateway/first_gateway/apiserver/routes/openai/deployments.py
bcote-anl: Great plan! I added the ClassVar implementation and got rid of the Enum class. Much cleaner. Thanks!

packages/gateway/first_gateway/apiserver/routes/openai/federated.py
bcote-anl: Done as above.

packages/gateway/first_gateway/apiserver/api.py
bcote-anl: Thanks for thinking and weighting options here, I appreciate it! Let's go with the current swap callback option for now. I'm wondering, however, if I should implement a feature to force recreating a backend client if it brakes? I'm not sure if this can happen, but let's say the httpx client somehow looses the connection. If it happens, it might means that the backend would be unreachable for all subsequent requests. What if we had an httpx error catch that would attempt to clear and recreate the httpx client if needed? The main pattern would still be "swap callback", but with a "lazy(-on-demand)-load" if needed as a backup plan. The codes might become a bit messier, however. Any thoughts? Maybe best for a future PR?

packages/gateway/first_gateway/apiserver/api.py
bcote-anl: Thanks :)

packages/gateway/first_gateway/apiserver/router_config_manager.py
bcote-anl: I agree, and thanks for your thoughts on this

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Sounds good. I incorporated a `self._configs` variable to capture the original BackendConfig and DeploymentConfig of each backend. If the code detects any change in one of them, it will close the client and force a re-creation. It's probably overkill to keep all of BackendConfig and DeploymentConfig, but at least it's a "clean" way to generalize the code and keep it minimal for the eventuality that the BackendConfig and DeploymentConfig definitions change in the future.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Done. We are now implementing 60 second sleep when a backend disapears. However, in the case of a backend client update (e.g. if something changes in the config like API key), I enforce a sleep of 0 seconds to make sure the change is implemented as quickly as possible. To remain flexible in the codes, I simply added the sleep duration as an input argument to the _close_backend(...) function. Hopefully this does not make the code too clunky ...

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Done. It's way better that way, of course! I kept the time to 2 years for now. I also think the same as you that we would likely restart our API at least once per 2 years. But feel free to add an issue so that we don't loose track if this point :)

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Yes, definitely a good time. For now I added the missing configs as global variables of the backend_client_manager.py file. But we can definitely make them input argument of the BackendClientManager class of env variable if it's more appropriate. I'm also grabbing max_concurrencies from Deployment's max_connections value. Happy to revise if more needs to be added.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: Sure, no problem :) Done.

packages/gateway/first_gateway/apiserver/api.py
masalim2: I think the httpx client should thankfully do most of this work that you're describing.  It internally manages a pool of open connections and recycles closed/broken connections automatically.  I wouldn't worry about this fallback method until we hit a real-life issue with the client!

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,134 @@
...
+            for backend in deployment.backends
+        }
+
+        # Backends that should be added or updated
+        for backend_id, (backend, deployment) in incoming.items():
+            if backend_id in self._clients and (
+                backend,
+                deployment,
+            ) != self._configs.get(backend_id):
+                await self._close_client(backend_id, sleep=0)
```
masalim2, line 93:
```suggestion
                logger.info(f"Closing client for {backend_id=}: configuration updated")
                await self._close_client(backend_id, sleep=0)
```

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,134 @@
...
+            if backend_id not in self._clients:
+                self._clients[backend_id] = self._create_client(backend, deployment)
+                self._configs[backend_id] = (backend, deployment)
+
+        # Backends that should be removed
+        for backend_id in set(self._clients) - incoming.keys():
+            await self._close_client(backend_id, sleep=60)
```
masalim2:
```suggestion
            logger.info(f"Scheduled _close_client for {backend_id=}: backend removed")
            await self._close_client(backend_id, sleep=60)
```

packages/gateway/first_gateway/apiserver/backend_client_manager.py
```diff
@@ -0,0 +1,134 @@
...
+
+    async def _close_client(self, backend_id: str, sleep: int = 0) -> None:
+        await asyncio.sleep(sleep)
+        client = self._clients.pop(backend_id)
```
masalim2:
```suggestion
        client = self._clients.pop(backend_id, None)
```

I'm pretty sure `pop()` will throw `KeyError` if it's missing without a default; we can add a `None` just in case to make it safe to retry

packages/gateway/first_gateway/apiserver/backend_client_manager.py
masalim2:
```suggestion
            task = asyncio.create_task(self._close_client(backend_id, sleep=60))
            _background_tasks.add(task)
            task.add_done_callback(_on_done)
```

If there are many backends to remove, let's run them all in parallel background tasks; as implemented, it will wait 60s per backend sequentially.  `_background_tasks` can be a module-scoped `_background_tasks: set[asyncio.Task[None]] = set()` and `_on_done` should do:

```python
def _on_done(task: asyncio.Task[None]) -> None:
    _background_tasks.discard(task)
    if task.cancelled():
        return
    if exc := task.exception():
        logger.error("Background _close_client failed", exc_info=exc)
```

packages/gateway/first_gateway/apiserver/backend_client_manager.py
masalim2: Looks great! Just added a followup comment on the background aspect (so multiple delayed cleanups happen after one 60s wait instead of running sequentially over 180s)

packages/gateway/first_gateway/apiserver/backend_client_manager.py
masalim2: Just in case some weird issue causes the config comparison to repeatedly close/recreate client connections: it would be good to surface this in the logs so we can be sure when clients are getting closed

packages/gateway/first_gateway/apiserver/backend_client_manager.py
masalim2: I agree with keeping it all.  It's cheap enough in memory and is flexible for changing config definitions, as you said.  One thing we will probably need to generalize in the future is the condition for recreating the clients on changes.  If something in the load balancing parameters like `weight` changes on a busy model, that could close some active connections.  It's probably only necessary to re-create the client on a handful of attributes that could be factored out into some `_should_recreate()` helper function.  But this is definitely an edge case for now: we can leave it as a todo.

bcote-anl: I see your point. I added `_should_recreate()` to check for changes in api_key, kind, max_backend_concurrency, and model_url.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Yes. Done.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Done. I applied the nomenclature to the previous logger.info call from the previous comment.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Done.

packages/gateway/first_gateway/apiserver/backend_client_manager.py
bcote-anl: Great plan. Done.

bcote-anl:
Hi @masalim2 I'm wondering if the reason `make test` is because the SSL cert is not on GitHub for the CI tests? `make test` worked on my laptop...

masalim2:
> Hi @masalim2 I'm wondering if the reason `make test` is because the SSL cert is not on GitHub for the CI tests? `make test` worked on my laptop...

Yes, you're right!  Pushed a fix by adding the step "Generate test CA" to `.github/workflows/ci.yml`.  This will generate a real temporary CA certificate and insert it into the necessary environment variables.

The missing certificates were not needed until the commit [moved ctx block to the __init__ function](https://github.com/argonne-lcf/FIRST/pull/307/commits/ac175e0c0ee8cda163ba089c926e1160ce3aef27).  This is because there's a dummy environment variable in ci.yml (`FIRST_PILOT_CA_CRT: "certificate"`) that allows the settings to load but was never actually used in the tests. 

After adding the `openssl` step to `BackendClientManager.__init__`, we now generate a cert at application startup, which triggered some of the test failures.

PR #317, opened 2026-08-25
re-organizing/cleaning dependencies.py

bcote-anl:
Reorganizing dependencies.py so that each Annotated call comes right after its function definition. 

I though of adding all function definitions first, followed by all Annotated calls, but since dependencies sometime depend on each other, it was not possible

Having "function/Annotated" blocks also group things nicely, since you can now see the function associated with each dependency more clearly.


masalim2 (approved):
Looks clean!

PR #318, opened 2026-08-25
Feature/estimate tokens

masalim2:
- Add `BasePayload.estimate_tokens()` with LLM payload implementations for OpenAI Chat Completions, OpenAI Responses, Anthropic Messages, OpenAI Embeddings.
- Add Anthropic Messages `BasePayload` subclass  for token estimation implementation (API route not implemented)
- Add implementations to estimate input tokens cheaply based on input prompt
    - Take care to avoid overcounting tokens from base64 encoded images
    - Take care to count tokens from large tool definitions
- Estimate output tokens based on modest cap and maximum context window
- Store maximum context window on Model, so users can query capability and configure agents correctly, and token estimation does not exceed the total available context


bcote-anl (approved):
I like how it's encoded within the pydantic payload models. Looks great!

bcote-anl:
I think that's a nice addition. You can merge whenever you are ready! We can improve as needed in the future. 
One idea could be to keep track of the input_token/output_token ratio from real previous requests to see how accurate the prediction was. But that's probably overkill at this stage...

PR #320, opened 2026-08-26
Extract token usage

### LINKED ISSUE #299: Extract Token Usage
Implement a token usage parser that supports 3 dialects (Chat Completions, Responses, Messages) x 2 modes (streaming/not-streaming).  This should extract the token usage at the end of each request and pass the usage data into the AdmissionController `settle()` method as well as emit a structured log (the v1 `RequestMetrics`).

There is a _totally unvetted_ but promising set of ideas in the AI-generated file https://github.com/argonne-lcf/FIRST/blob/firstv2/docs/architecture/usage-adapters.py

masalim2:
Added support to extract `TokenUsage` from: 

- chat/completions (streaming + unary)
- responses (streaming + unary)
- messages (streaming + unary)
- embeddings (unary)

The unary case is wired into `submit_inference_with_retry()`.  The streaming case should be integrated in the future with:

```python
tap = UsageTap()
async for chunk in resp.aiter_raw():
    tap.feed(chunk)
    yield chunk

tap.close()
usage = USAGE_PARSERS[payload.endpoint].parse_stream(tap.first, tap.last)
```
 
I created real example responses for each LLM dialect in both streaming and unary variants in `tests/sample_prompts/`.  These were created from vllm on Minerva.  `test_usage.py` then verifies that the parsers work correctly on each response format.  

Closes #299


bcote-anl (approved):
I like the clean approach of having a usage.py file to manage all of the token parsing mechanics!

The incorporation into submit_inference.py is pretty light and minimal, which is great!

PR #321, opened 2026-08-26
Renew inflight requests in AdmissionController

### LINKED ISSUE #298: Refresh in-flight request IDs in the admission controller
Each api worker should maintain a set of _currently-active_ in-flight request IDs.  These are requests that have been successfully admitted by the AdmissionController and not yet settled.  Can use `set.add` & `set.remove` to efficiently maintain the set during `admit` and `settle`.  

Run a periodic background task (once per 10 seconds) that passes the in-flight request IDs to `AdmissionController.renew()`.  This renews the expiration of reserved slots in the admission controller for healthy API workers.  API workers that crash will then naturally stop renewing, and the reserved slots will get freed.

The background loop and data structure should be established in the app lifespan.  We could consider adding the functionality directly to the `AdmissionController` class and then maintaining a persistant on `app.state` instead of constructing a new instance per request.  Alternatively, keep the `AdmissionController` an ephemeral dependency and attach a separate helper to the app that maintains the set and background loop.

masalim2:
This is adds a necessary final component to the AdmissionController without impacting the usage in the inference APIs.  Instead of creating a new `AdmissionController()` per request, a single instance is maintained on `app.state`.  

The instance can then keep track of all the inflight requests per uvicorn process.  This inflight set is passed to `renew()` every 10 seconds to extend the lease on inference requests that are still running.  

This prevents the sweeper from incorrectly expiring inflight requests that are still truly running.  

[Context: The sweeper, in turn, ensures that requests on crashed workers are correctly cleaned up: worker dies, doesn't renew lease for 30 seconds, sweeper removes the dead reservations from Redis, returning the model/user concurrency counts back to the correct level.]

### DIFF
--- a/packages/gateway/first_gateway/apiserver/api.py
+++ b/packages/gateway/first_gateway/apiserver/api.py
@@ -6,6 +6,7 @@ from fastapi import FastAPI
 
 from first_gateway.settings import Settings
 
+from ..database.redis.admission import AdmissionController
 from ..log_config import config_logging
 from .backend_client_manager import BackendClientManager
 from .error_handlers import register_error_handlers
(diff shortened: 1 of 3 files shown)


bcote-anl (approved):
LGTM!

Issue #294, opened 2026-08-20
Generalize submit_inference_with_retry for streaming


masalim2:
- The current implementation of `submit_inference_with_retry` with `finally: settle` is robust for non-streaming requests, but streaming settlement will have to occur inside of the async generator returned by the `StreamingResponse`. 
- Generalize the implementation for both non-streaming and streaming request pathways.

See the rough sketch in https://github.com/argonne-lcf/FIRST/blob/firstv2/docs/architecture/nginx-httpx.md#streaming-requests.  The `finally` block inside the relay generator would handle the `settle` cleanup and streaming finalization

Issue #311, opened 2026-08-24
Refactor around retryable inference loop


masalim2:
# API Inference Request Code Organization

In other PRs we had mentioned the need to generalize the inference request path.  Some preliminary thoughts on separating code by the stage: 

## Before Retry Loop

This is already in pretty good shape.  This code should generalize for all inference endpoints (LLM/non-LLM) and is mostly the work that needs to be done before admitting the request:

- Authorize user-model access
- Generate request ID
- Estimate tokens from Payload.estimate_tokens() (`BasePayload.estimate_tokens()` defaults to 0; set an estimation for each LLM subclass)
- Select/shuffle candidate backends by weight

## Retry Loop

There is a good chunk of logic at this layer which can also generalize across inference endpoints (not specific to the endpoint, model, or stream parameters):

- Select backend and admit request
- Get configured AsyncClient for selected backend
- Handoff to specific **upstream handler**: Request ID, Async Client, payload
- Return upstream handler's response to user
- If the upstream handler raises an error, handle it here (possibly retrying, if the error type is retryable)

## The upstream handler
To make the retrying inference loop generic enough, we would separate out an **upstream handler** that's responsible for:
    
- Returning a response or raising an error
- Settling/cleanup (finally closing the upstream connection and settling with admission controller)
- Response introspection (e.g. token usage data extraction)
- Emitting post-response events (e.g. Request Metrics Logs)

Depending on the types of `FirstError` that are raised by the upstream handler, we can take different actions: 

- Propagate full error directly back to user; do not retry (a 4xx from vLLM is safe to relay and is probably their fault)
- Immediate failure; send our own message back to user (not retryable; don't want to surface internal problem to user)
- Retryable failure; can try another backend

## Upstream handlers 

- Non-streaming LLM request
- Streaming LLM request
- Disaggregated LLM request (internally orchestrates 2 requests for prefill->decode flow)
- Generic synchronous REST endpoint (
- Generic async REST endpoint
    - Send a JSON body in a background task; return 202 ACCEPTED immediately to user with a task_id they can poll to retrieve result.  Result relayed through redis.
- Generic sync gRPC endpoint

Issue #302, opened 2026-08-20
Refine backend error handling


masalim2:
If every backend fails a request attempt, we return this to the user:

```python
raise ServiceUnavailable(f"Too many failed attempts on backends for model {model}.")
```

After the inference routes are implemented, we should revisit the error handling logic such that:

- a `4xx` error from a backend can safely be propagated back to the user without retrying on other backends: something is probably wrong with their request.  Expose the error body because it may contain the useful API validation error from vLLM.  
- a `5xx` status error or any subclass of `httpx.RequestError` can be retried on another backend without relaying the internal error to the user

If all backend attempts fail, raise the opaque `ServiceUnavailable`

PR #322, opened 2026-08-27
Firstv2 streaming

bcote-anl:
Adding streaming option for submitting inference with retry. I am having a hard time keeping things clean, since I had to hardcode the try/finally block inside the async iterator. But now we have two self-contained functions to submit inference: submit_inference_streaming and submit_inference.

I was not able to test since my job is stuck in the queue for Tara. I'll make a test tomorrow, but @masalim2 I wanted you to be able to see where this is going to accelerate the process.

One little thing, in terms of generalizing this, should I rename the following?
submit_inference_streaming --> submit_streaming
submit_inference --> submit_unary

Thanks in advanced for the review!

Would close issue #294

### DIFF
--- a/packages/gateway/first_gateway/services/submit_inference.py
+++ b/packages/gateway/first_gateway/services/submit_inference.py
@@ -40,7 +40,7 @@ async def submit_inference_with_retry(
 
     backend_candidates = get_backend_candidates(model, deployment_name=deployment_name)
     if not backend_candidates:
-        raise ServiceUnavailable(f"No backend available for model {model}.")
+        raise ServiceUnavailable(f"No backend available for model {model.name}.")
 
     estimated_tokens = payload.estimate_tokens(model.max_model_len)
     streaming = getattr(payload, "stream", False) or False
--- a/packages/gateway/first_gateway/services/submit_inference.py
+++ b/packages/gateway/first_gateway/services/submit_inference.py
@@ -67,6 +67,8 @@ async def submit_inference_with_retry(
                     payload,
                     admission_controller,
                     request_id,
+                    user,
+                    model,
                 )
             else:
                 response = await submit_inference(
@@ -74,6 +76,8 @@ async def submit_inference_with_retry(
                     payload,
                     admission_controller,
                     request_id,
+                    user,
+                    model,
                 )
         except:
             # TODO: figure out a way to gather and report errors
@@ -98,6 +102,8 @@ async def submit_inference(
     payload: BasePayload,
     admission_controller: AdmissionController,
     request_id: str,
+    user: AuthUser,
+    model: ModelConfig,
 ) -> dict[str, Any]:
     """POST to an inference backend."""
 
@@ -110,6 +116,8 @@ async def submit_inference(
     body = cast(dict[str, Any], response.json())
     parser = USAGE_PARSERS.get(payload.endpoint)
     usage = parser.parse_unary(body) if parser else TokenUsage()
+    # TODO: emit structured log events (this is a placeholder for visibility):
+    logger.info(f"{payload.endpoint} - {model.name} - {user.username} - {usage}")
     await admission_controller.settle(request_id, actual_tokens=usage.total_tokens or 0)
     return body
 
@@ -119,6 +127,8 @@ async def submit_inference_streaming(
     payload: BasePayload,
     admission_controller: AdmissionController,
     request_id: str,
+    user: AuthUser,
+    model: ModelConfig,
 ) -> StreamingResponse:
     """POST to an inference backend and relay the SSE stream to the caller."""
 
@@ -145,6 +155,9 @@ async def submit_inference_streaming(
                 await response.aclose()
             usage = parser.parse_stream(tap.first, tap.last) if parser else TokenUsage()
             total_tokens = usage.total_tokens or 0
+            logger.info(
+                f"{payload.endpoint} - {model.name} - {user.username} - {usage}"
+            )
             await admission_controller.settle(request_id, actual_tokens=total_tokens)
 
     return StreamingResponse(
--- a/packages/gateway/first_gateway/services/submit_inference.py
+++ b/packages/gateway/first_gateway/services/submit_inference.py
@@ -1,7 +1,9 @@
 import logging
-from typing import Any, cast
+from typing import Any, AsyncIterator, cast
 
+import anyio
 from fastapi.responses import StreamingResponse
+from httpx import AsyncClient
 
 from first_common.errors import ServiceUnavailable
 from first_common.schema.endpoints.base import BasePayload
@@ -15,7 +17,7 @@ from .orchestration import (
     get_backend_candidates,
     get_deployment_from_backend_id,
 )
-from .usage import USAGE_PARSERS, TokenUsage
+from .usage import USAGE_PARSERS, TokenUsage, UsageTap
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -51,31 +54,42 @@ async def submit_inference_with_retry(
...
+                )
         except:
             # TODO: figure out a way to gather and report errors
             deployment = get_deployment_from_backend_id(model.deployments, backend_id)
             await admission_controller.record_error(
                 backend_id, deployment.router_params
             )
+            await admission_controller.settle(request_id, actual_tokens=0)
             backend_candidates = [b for b in backend_candidates if b.uid != backend_id]
             logger.warning(f"Backend {backend_id} failed, trying next one.")
```
masalim2:
```suggestion
            logger.error(f"Backend {backend_id} failed, trying next one.", exc_info=True)
```

Let's tackle #302 for full error handling in a followup PR.  But for now, we can raise the severity and surface the exception in the logs to help with debugging if we hit `Too many failed attempts on backends`

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -38,9 +40,10 @@ async def submit_inference_with_retry(
 
     backend_candidates = get_backend_candidates(model, deployment_name=deployment_name)
     if not backend_candidates:
-        raise ServiceUnavailable(f"No backend available for model {model}.")
+        raise ServiceUnavailable(f"No backend available for model {model.name}.")
```
masalim2: This line just reminded me of another important detail: calling `AdmissionController.admit()` and receiving a result of `AdmitStatus.REJECT_CAPACITY` provides the necessary signal to the AutoScaler to cold-start a model.  

If there are no backend candidates, this early-fail logic makes sense, but it will skip sending the signal to Redis that the model needs to scale up.   Your `raise_admit_error()` function actually already handles the no backends available scenario:

```python
 elif admit_result.capacity_reason == CapacityReason.NO_CANDIDATES:
            raise ServiceUnavailable(
                "Backend not available yet." + retry_str,
                retry_after_sec=retry_after_sec,
            )
```

Can we slightly restructure this function to be agnostic to `len(backend_candidates)` and always attempt at least one loop iteration of `admit_request()`, even if the candidate list is empty?  When it re-builds `backend_candidates` during request exception handling, then it can `break` if the list is empty (but we always at least try once).  

That should actually be a nice small simplification too, since the "no backends" error is handled in one place.

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -51,31 +54,42 @@ async def submit_inference_with_retry(
             request_id,
             estimated_tokens=estimated_tokens,
         )
-        total_tokens = 0
+
+        client = backend_client_manager.get(backend_id)
+        if client is None:
+            raise ServiceUnavailable(f"No client exist for backend ID {backend_id}.")
```
masalim2:
```suggestion
            await admission_controller.settle(request_id, actual_tokens=0)
            raise ServiceUnavailable(f"No client exist for backend ID {backend_id}.")
```

- We should call `settle()` here to prevent a reservation leak if we ever hit this unexpected error.  

---

I also started thinking a little more about if this error case is ever possible: I think it's going to be quite rare and basically impossible while a deployment is at a fixed number of replicas or the traffic isn't so busy that multiple requests are queued on a uvicorn worker at a time.  

However, imagine:  a model is getting blasted by users during a busy session, that triggers the autoscaler to start a new replica, and the replica eventually becomes healthy and calls `RouterConfigManager.swap()`.  During that busy window of time, the asyncio runtime is interleaving between different requests and your background swap callbacks at every `await` point.  It's possible that:

1. The new healthy backend_id is added to the `RouterConfig` manager
2. It calls `await` on the swap callback to update the `BackendClientManager`.
3. Execution jumps over to a new request for the same model, which picks up the new healthy backend ID and selects it.
4. We get to this point in the code where the Backend ID exists in the Router Config but has not yet been added to the client manager.

So I think this is a possible race condition that would cause us to hit this exception, but I do think it can only be observed if the system is very busy in the middle of a scaling event.  We can add a simple layer of defense by filtering `backend_candidates` down to only those that exist in `BackendClientManager`:

```python
backend_candidates = get_backend_candidates(model, deployment_name=deployment_name)
backend_candidates = [b for b in backend_candidates if b.uid in backend_client_manager]
```

I think the other possible race condition, that a client is yanked out in the middle of a request, is already well-covered by the 60 second sleep before client deletion.

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -84,20 +98,70 @@ async def submit_inference_with_retry(
...
+    client: AsyncClient,
+    payload: BasePayload,
+    admission_controller: AdmissionController,
+    request_id: str,
+    user: AuthUser,
+    model: ModelConfig,
+) -> StreamingResponse:
+    """POST to an inference backend and relay the SSE stream to the caller."""
+
+    request = client.build_request(
```
masalim2, line 148:
```suggestion
    payload.stream = True
    request = client.build_request(
```

Just in case we ever make `stream = True` the default in the future: `exclude_unset` will drop defaults that the user did not specify, and this is just a small safeguard to make sure the request body is consistent with the streaming request

packages/gateway/first_gateway/services/submit_inference.py
```diff
@@ -84,20 +98,70 @@ async def submit_inference_with_retry(
...
+        "POST",
+        f"/v1/{payload.endpoint}",
+        json=payload.model_dump(exclude_unset=True, mode="json"),
+    )
+
+    response = await client.send(request, stream=True)
+    response.raise_for_status()
```
masalim2:
```suggestion
    if response.status_code != 200:
        body = await response.aread()
        await response.aclose()
        # TODO: streaming error handling
        logger.warning(f"Received error from backend: {body[-512:]}")
        response.raise_for_status()
```

masalim2 (commented):
This is awesome, thank you so much!

> I am having a hard time keeping things clean, since I had to hardcode the try/finally block inside the async iterator. ...  One little thing, in terms of generalizing this, should I rename the following...

I wouldn't worry about these issues for now, you've honestly done an amazing job at keeping each changeset focused, easy to review, with a strong progression of functionality in the sequence of PRs.  In that spirit, I think this PR's scope nails it with a self-contained streaming function that runs under the same parent retry loop.  

In followup PRs we can flesh out the error handling a little bit more fully (#302) and do some light refactoring around the inference functions and where the cleanup/settle logic lives (#311). I've honestly always been guilty of scope creep in my own PRs and am trying to follow your lead in keeping change sets focused, because the PR review experience has been great.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: I like it! Done.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: Yes, great plan, I edited the loop to at least go in once. Not I can see the error raised from `raise_admit_error()`. Thanks!

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: I see what you mean, I implemented it thinking of this race case. But I like you suggestion of filtering `backend_candidates`, it's much more elegant and reliable. I added the feature!

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: Agreed. I added it.

packages/gateway/first_gateway/services/submit_inference.py
bcote-anl: Sounds good! Added.

